PR #7, opened 2014-01-30
Adding Katello repository support to allow installation of the Katello

ehelms:
RPMs if not already installed on the system.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -2,10 +2,9 @@
 class katello::config {
   include katello::params
 
-  group { $katello::group:
+  group { $katello::user_groups:
     ensure => "present"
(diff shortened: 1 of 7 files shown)


ekohl:
Is this still relevant? Needs a serious rebase.

ehelms:
@ekohl Solid question, I can re-tackle this if we think it worth the effort of the module handling this.

ekohl:
@ehelms There's (at least) one use case that can benefit from the module managing the repositories and that's using the foreman to roll out additional Katello instances. I don't know how common that use case is. Given the age of this PR I'd now close it and only reconsider it if the need arises again (or PRs).

PR #50, opened 2014-12-04
Fixes #8585: Remove unused configuration and unneeded functions.

ehelms:


stbenjam:
ACK

ekohl:
Same as #64, I think you can drop ```foreman_oauth_key``` from ```params.pp``` while you're at it (unless some other module secretly uses it, but I doubt that).

templates/katello.yml.erb
```diff
@@ -12,43 +12,24 @@
...
-  port:
-  use_ssl: true
-  use_foreman: <%= scope.lookupvar("katello::params::use_foreman") %>
-
-  post_sync_url: https://localhost<%= scope.lookupvar("katello::params::deployment_url") %>/api/v2/repositories/sync_complete?token=<%= scope.lookupvar("post_sync_token") %>
+  post_sync_url: https://localhost<%= @deployment_url %>/api/v2/repositories/sync_complete?token=<%= @post_sync_token %>
 
   candlepin:
-    url: <%= scope.lookupvar("katello::params::candlepin_url") %>
+    url: <%= @candlepin_url %>
```
ekohl: I'm slightly surprised this still works with scoping. I'd recommend adding a test to ```katello_config_spec.rb``` that verifies one of the config vars (but possibly all).

ehelms: I'd be ok with changing all of these to `scope[]` syntax if that would alleviate the potential for strangeness in the future.

ekohl: https://docs.puppetlabs.com/guides/templating.html#referencing-variables mentions that they should be in scope and since the template is rendered from ```katello::config```, I'd *expect* everything from the ```katello``` scope to be unavailable.

ehelms:
@ekohl Added tests.

ekohl:
@ehelms since there's test coverage now, I assume it's ok. Once the aim is puppet 4 support, it can just be added to the test matrix and you'll find out easily enough. :+1:

stbenjam:
ACK

PR #54, opened 2015-01-07
Fixes #8849 - installs foreman_gutterball

dustints:


komidore64:
didn't need to remove anything wrt the gutterball backend service?

ehelms:
Why not include calling this in the section of init.pp that sets up gutterball? Is there ever an instance where we want to install and configure gutterball but omit the plugin for it?  Or vice versa, move the configuration of gutterball into the plugin's manifest file such that if gutterball is being setup and configured the plugin does as well and then it's all nicely wrapped up in a manifest file.

manifests/init.pp
```diff
@@ -116,9 +116,12 @@
 
   Service['httpd'] -> Exec['foreman-rake-db:seed']
 
+  class { 'katello::plugin': }
+
   if $katello::gutterball {
```
ehelms: Can we move this chunk and logic in plugin/gutterball.pp ? Just makes more sense to all live together for me.

dustints: @ehelms removed the extra logic.

ehelms:
I think I am getting lost in translation. I'd suggest putting all the gutterball logic in katello::plugins::gutterball and then either:

1) In init.pp, include only:
```
if $katello::gutterball {
  class { 'katello::plugins::gutterball': }
}
```

2) Use the installers answer file to set the plugin to true when it needs to be included and treat gutterball integration as a true plugin similar to other Foreman plugins:

```
katello::plugins::gutterball: true
```

These options let gutterball and it's integration be more cohesive and less coupled.

manifests/plugin/gutterball.pp
```diff
@@ -0,0 +1,9 @@
+# gutterball plugin
+class katello::plugin::gutterball{
+    Class[ 'certs' ] ->
```
ehelms: Nitpick - indentation is off with respect to other files (2 spaces instead of 4)

manifests/plugin/gutterball.pp
```diff
@@ -0,0 +1,9 @@
...
+    class { 'certs::gutterball': } ->
+    foreman::plugin { 'gutterball': } ->
+    class { 'gutterball':
```
ehelms: This will need to be `class { '::gutterball':` to prevent duplicate resource error. I think a simple spec test of this plugin manifest might be nice as it would have caught this error without having to run a full katello-installer test run.

dustints:
@ehelms added the spec but running into some issues getting a passing spec...

foreman isn't being evaluated 
so foreman::plugin_prefix used in the foreman plugin type is empty causing it to install gutterball package
but gutterball module also installs the gutterball package.
so getting duplicate declaration of package 'gutterball'

do you know any way around this issue?

```
 [[1;31mWarning: Scope(Foreman::Plugin[gutterball]): Could not look up qualified variable 'foreman::plugin_prefix'; class foreman has not been evaluat»
^[[0;36mDebug: importing '/home/dustin/puppet-katello/spec/fixtures/modules/foreman/manifests/service.pp' in environment production^[[0m¬
^[[0;36mDebug: Automatically imported foreman::service from foreman/service into production^[[0m¬
^[[1;31mError: Duplicate declaration: Package[gutterball] is already declared in file /home/dustin/puppet-katello/spec/fixtures/modules/gutterball/man»
F¬                                                                                                                                                     
¬
Failures:¬
¬
  1) katello::plugin::gutterball should contain Class[certs]¬
     Failure/Error: it { should contain_class('certs') }¬
     Puppet::Error:¬
       Duplicate declaration: Package[gutterball] is already declared in file /home/dustin/puppet-katello/spec/fixtures/modules/gutterball/manifests/i»
     # ./spec/classes/plugin_gutterball_spec.rb:14:in `block (2 levels) in <top (required)>'¬
¬
  2) katello::plugin::gutterball should call the plugin¬
     Failure/Error: should contain_foreman__plugin('gutterball')¬
     Puppet::Error:¬
       Duplicate declaration: Package[gutterball] is already declared in file /home/dustin/puppet-katello/spec/fixtures/modules/gutterball/manifests/i»
     # ./spec/classes/plugin_gutterball_spec.rb:18:in `block (2 levels) in <top (required)>'¬
¬
```

spec/classes/plugin_gutterball_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
+
+describe 'katello::plugin::gutterball' do
+  let(:foreman) { {:plugin_prefix => 'ruby193-'} }
```
ekohl: Does this work? I think we mostly use pre_condition:
```ruby
let :pre_condition do
  "class {'foreman': plugin_prefix => 'ruby-193-'}"
end
```

Though in this particular case you don't have to specify the plugin_prefix explicitly since it should detect it from the provided facts.

spec/classes/plugin_gutterball_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let :facts do
+    {
+      :concat_basedir             => '/tmp',
+      :operatingsystem            => 'RedHat',
+      :operatingsystemrelease     => '6.4',
+      :operatingsystemmajrelease  => '6.4',
```
ekohl: This should be '6' since it's only the major version.

dustints:
thanks @ekohl, that worked!

ehelms:
ACK

PR #55, opened 2015-01-09
Fixes #8849 - fix plugin gutterball for katello-devel-installer usage

dustints:


ehelms:
I think the devel module itself should handle setting up gutterball if it needs to - given that the devel setup won't load or use the RPM version of foreman_gutterball (and instead would use the gem version via a bundler file), configuring this module to account for that doesn't make sense to me.

dustints:
@ehelms ah gotcha...abandoning

PR #56, opened 2015-01-20
Ref #9055 - make packages depended on, a configurable parameter

dustints:

PR #57, opened 2015-01-26
Ref #9055 - param to depend on packages that may have an scl prefix

dustints:


manifests/params.pp
```diff
@@ -14,7 +14,9 @@
         }
       }
 
-      $package_names = ['katello', "${scl_prefix}rubygem-katello"]
+      $scl_package_names = ['rubygem-katello']
+      $scl_package_names_with_prefix = $scl_package_names.map |$pkg| { "${scl_prefix}${pkg}" }
```
stbenjam: `map` is relatively new - 3.4 I think, yet we're testing with even 2.7 on travis.  stdlib includes a `prefix` function which might be a better option.

ehelms:
The problem with this would be, in the installers case, having to configure answers files per dist unless you moved the scl_package_names_with_prefix logic into init.pp where it can be evaluated after the parameters have been supplied.

ekohl:
What's the added benefit of this? The installer answer file is a good case I didn't consider for https://github.com/Katello/puppet-katello/pull/56

dustints:
@ehelms updated,... i wasn't able to use the map function without changing the puppet.conf file...so decided to use a template to achieve the same thing (feels really hacky).  I moved scl_package_names_with_prefix into install.pp

@ekohl would like to include rubygems-katello as a dependency for sam-installer in https://github.com/Katello/puppet-katello/pull/56 but there is the posibility that some distros need the scl-prefix like ''ruby-193-rubygems-katello" for the package name.

manifests/install.pp
```diff
@@ -1,6 +1,13 @@
 # Katello Install
 class katello::install {
+  $scl_prefix = $katello::scl_prefix
+  $scl_package_names = $katello::scl_package_names
+  $scl_package_names_with_prefix = split(template('katello/sclpackages.erb'), ',')
```
ekohl: @stbenjam suggested prefix from stdlib: ```prefix($scl_package_names, $scl_prefix)``` does the same thing.

ekohl:
I'm not sure why the current logic doesn't suffice for that. It already uses non-SCL for Fedora and the ruby193 SCL for other RH-based distros. What makes the sam-installer special that it needs to override the package names

dustints:
looks like we don't need it anymore..based on comments on https://github.com/Katello/puppet-katello/pull/56

PR #58, opened 2015-01-28
Ref #9134 - gb plugin sets configs for foreman-gutterball

dustints:
depends on https://github.com/Katello/puppet-gutterball/pull/6

### DIFF
--- a/manifests/plugin/gutterball.pp
+++ b/manifests/plugin/gutterball.pp
@@ -1,9 +1,15 @@
 # gutterball plugin
 class katello::plugin::gutterball{
+
   Class[ 'certs' ] ->
   class { 'certs::gutterball': } ->
   foreman::plugin { 'gutterball': } ->
   class { '::gutterball':
     keystore_password => $certs::gutterball::gutterball_keystore_password,
+  } ~>
+  class { 'katello::plugin::gutterball::config':
+    foreman_plugins_dir => $katello::config_dir,
+    foreman_user        => $katello::user,
+    foreman_group       => $katello::group,
   }
 }
--- a/manifests/plugin/gutterball/config.pp
+++ b/manifests/plugin/gutterball/config.pp
@@ -1,0 +1,15 @@
+# gutterball config
+class katello::plugin::gutterball::config(
+    $foreman_user  = 'foreman',
+    $foreman_group = 'foreman',
+    $foreman_plugins_dir = '/etc/foreman/plugins',
+    $url = "https://${::fqdn}:8443/gutterball",
+  ){
+
+  file { "${foreman_plugins_dir}/gutterball.yaml":
+    content => template('katello/plugin/gutterball/gutterball.yaml.erb'),
+    owner   => $foreman_user,
+    group   => $foreman_group,
+    mode    => '0644',
+  }
+}
--- a/templates/plugin/gutterball/gutterball.yaml.erb
+++ b/templates/plugin/gutterball/gutterball.yaml.erb
@@ -1,0 +1,2 @@
+foreman_gutterball:
+  url: <%= @url %>


templates/foreman_gutterball.yaml.erb
```diff
@@ -0,0 +1,2 @@
+<%# for foreman_gutterball plugin %>
+:url: <%= @fqdn %>
```
komidore64: this should have `/gutterball` on the end. other than that, i will see if i can test this on monday!

ehelms:
A more tpyical convention would be to create `manifests/plugin/gutterball/config.pp` and perhaps to mimic the same for templates to help with separation and locating things, e.g. `templates/plugin/gutterball/foreman_gutterball.yaml.erb`

manifests/plugin/foreman_gutterball_config.pp
```diff
@@ -0,0 +1,28 @@
+# foreman_gutterball config
+class katello::plugin::foreman_gutterball_config(
+    $foreman_user  = 'foreman',
+    $foreman_group = 'foreman',
+    $foreman_plugins_dir = '/usr/share/foreman/plugins',
```
ehelms: I don't think this is the right location for the file to get laid down. I believe it should be `/etc/foreman/settings.plugins.d/` . See https://github.com/theforeman/foreman/blob/develop/config/settings.rb#L14

templates/foreman_gutterball.yaml.erb
```diff
@@ -0,0 +1,2 @@
+<%# for foreman_gutterball plugin %>
```
ehelms: I would consider namespacing your settings, e.g.:

```
:foreman_gutterball:
  :url:
```

If you don't namespace then `url` gets put into the top level settings namespace and could be over written easily. This way you can access SETTINGS[:foreman_gutterball][:url] and it's clear what 'url' is related to and les slikely to be over written.

manifests/plugin/foreman_gutterball_config.pp
```diff
@@ -0,0 +1,28 @@
...
+    $foreman_gutterball_gem_config = undef
+  ){
+  $config_file = "${foreman_plugins_dir}/foreman_gutterball.yaml"
+
+
+  concat{ $config_file:
```
ehelms: I am not following why all this concatenation needs to occur to deploy this settings file?

manifests/plugin/foreman_gutterball_config.pp
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  concat::fragment { "foreman_gutterball.yaml from template":
+    target => $config_file,
+    content => template('katello/foreman_gutterball.yaml.erb'),
+    order => '01'
+  }
+  concat::fragment { "gem's foreman_gutterball.yaml":
+    target => $config_file,
+    source => "$foreman_gutterball_gem_config",
```
ekohl: Does not have to be quoted.

manifests/plugin/foreman_gutterball_config.pp
dustints: @ehelms hoping to allow creating settings from foreman-gutterball and here.

dustints: @ehelms i guess concat doesn't have to happen..just a new file name would work
...will address.

templates/foreman_gutterball.yaml.erb
dustints: @ehelms is it ok for url to fall into something like :foreman_gutterball_common:
i think if the 2 settings use the same namespace, one will overwrite the other instead of merging them.

templates/foreman_gutterball.yaml.erb
dustints: addressing

dustints:
@ehelms, @komidore64 , @ekohl 
addressed commends.
* template is now @ templates/plugin/gutterball/foreman_gutterball_addressable.yaml.erb
* does not concat with the gem's settings into one config. instead it creates a new config in /usr/foreman..settings.plugin.d/foreman_gutterball_auxiliary.yaml

manifests/plugin/gutterball.pp
```diff
@@ -1,9 +1,26 @@
 # gutterball plugin
 class katello::plugin::gutterball{
+  $foreman_gutterball_gem_config_tmp = '/tmp/foreman_gutterball.yaml'
+
+  $package = "${foreman::plugin_prefix}gutterball"
```
ekohl: This whole block appears to be unused.

templates/sclpackages.erb
```diff
@@ -0,0 +1 @@
+<%= @scl_package_names.map{ |pkg| @scl_prefix + pkg }.join(',') %>
```
ekohl: This appears to be unrelated.

manifests/plugin/gutterball/config.pp
```diff
@@ -0,0 +1,15 @@
+# gutterball config
+class katello::plugin::gutterball::config(
+    $foreman_user  = 'foreman',
+    $foreman_group = 'foreman',
+    $foreman_plugins_dir = '/usr/share/foreman/config/settings.plugins.d',
```
ekohl: This should be in /etc.

manifests/plugin/gutterball.pp
```diff
@@ -1,9 +1,26 @@
...
+    }
+  }
+
   Class[ 'certs' ] ->
   class { 'certs::gutterball': } ->
   foreman::plugin { 'gutterball': } ->
   class { '::gutterball':
     keystore_password => $certs::gutterball::gutterball_keystore_password,
+  } ~>
+  class { 'katello::plugin::gutterball::config':
```
ekohl: I think this doesn't have to be a separate class since it's only a single file. I even intend to make this possible in https://github.com/theforeman/puppet-foreman/pull/281 with a parameter to ```foreman::plugin```.

manifests/plugin/gutterball/config.pp
```diff
@@ -0,0 +1,15 @@
...
+    $foreman_gutterball_gem_config = undef
```
ekohl: This appears to be unused.

manifests/plugin/gutterball.pp
```diff
@@ -1,9 +1,26 @@
 # gutterball plugin
 class katello::plugin::gutterball{
+  $foreman_gutterball_gem_config_tmp = '/tmp/foreman_gutterball.yaml'
```
ekohl: This appears to be unused.

manifests/plugin/gutterball.pp
dustints: that would be nice to let configs manage their own configs easily.
I kept it in it's own class so it could be reused by puppet-katello_devel https://github.com/Katello/puppet-katello_devel/pull/30

dustints:
@ekohl  updated.

manifests/plugin/gutterball/config.pp
```diff
@@ -0,0 +1,14 @@
+# gutterball config
+class katello::plugin::gutterball::config(
+    $foreman_user  = 'foreman',
+    $foreman_group = 'foreman',
+    $foreman_plugins_dir = '/etc/foreman/plugins',
+  ){
+
+  file { "${foreman_plugins_dir}/foreman_gutterball_auxiliary.yaml":
```
ehelms: Why 'auxiliary' ?

manifests/plugin/gutterball/config.pp
```diff
@@ -0,0 +1,14 @@
+# gutterball config
+class katello::plugin::gutterball::config(
+    $foreman_user  = 'foreman',
+    $foreman_group = 'foreman',
+    $foreman_plugins_dir = '/etc/foreman/plugins',
```
ehelms: I still think this is the wrong directory - https://github.com/Katello/puppet-katello/pull/58#discussion-diff-23880066

templates/plugin/gutterball/foreman_gutterball_auxiliary.yaml.erb
```diff
@@ -0,0 +1,2 @@
+:foreman_gutterball_auxiliary:
+  :url: <%= @fqdn %>:8443/gutterball
```
ekohl: I think this should also be parameterized.

manifests/plugin/gutterball/config.pp
ekohl: It's not. 
```
# ls -l /usr/share/foreman/config/settings.plugins.d
lrwxrwxrwx. 1 root root 20 jan 29 18:38 /usr/share/foreman/config/settings.plugins.d -> /etc/foreman/plugins
```

manifests/plugin/gutterball/config.pp
dustints: what's a good name...forman_gutterball_url?

ehelms: How about just `foreman_gutterball.yaml` ? I don't see why it needs a special name when it is simply the config file for foreman_gutterball.

dustints:
@ehelms, @ekohl updated
* rename foreman_gutterball_aux => foreman_gutterball_url.yaml
* parameterize the url supplied in template.

manifests/plugin/gutterball/config.pp
ekohl: I'd use gutterball.yaml since the plugin is also named gutterball. In the PR I mentioned I will also default to that if you don't specify a config filename.

templates/plugin/gutterball/foreman_gutterball_url.yaml.erb
```diff
@@ -0,0 +1,6 @@
+:foreman_gutterball_url:
+  :protocol: <%= @protocol %>
```
ekohl: Does the plugin use these variables? I'd only parameterize the configuration parameters that the plugin actually uses.

dustints: @komidore64 do you think there is any value for foreman-gutterball to have discrete portions of the url?

manifests/plugin/gutterball/config.pp
komidore64: @ekohl how many dirs are there for plugin configs? what is the correct one to use?

templates/plugin/gutterball/foreman_gutterball_url.yaml.erb
komidore64: we could go either way. i was doing similar to what is found in katello's config

for example:
```yaml
common:
  # ...
  candlepin:
    url: https://localhost:8443/candlepin
```

dustints:
@komidore64, @ehelms, @ekohl 
updated
* just using the same format as katello's config for url
* let this module manage foreman_gutterball.yaml so renaming templates and destination config file to foreman_gutterball.yaml

manifests/plugin/gutterball/config.pp
```diff
@@ -0,0 +1,17 @@
...
+    $protocol = 'https',
+    $port = '8443',
+    $path = 'gutterball/',
```
ehelms: I think we typically defer to leading slashes instead of trailing slashes for URL paths.

manifests/plugin/gutterball/config.pp
komidore64: i have had `/etc/foreman/plugins/<title>.yaml` confirmed to me on IRC.

dustints:
@ehelms updated

ehelms:
APT

templates/plugin/gutterball/foreman_gutterball.yaml.erb
```diff
@@ -0,0 +1,2 @@
+:foreman_gutterball:
+  :url: <%= @protocol %>://<%= @fqdn %>:<%= @port %><%= @path %>
```
komidore64: one last thing...

let's stick with the keys being strings instead of symbols, so we can stay uniform with other configs.

so
```erb
foreman_gutterball:
  url: <%= @protocol %>://<%= @fqdn %>:<%= @port %><%= @path %>
```

dustints:
@komidore64 updated

komidore64:
awesome!

ack pending tests

manifests/plugin/gutterball/config.pp
```diff
@@ -0,0 +1,17 @@
+# gutterball config
+class katello::plugin::gutterball::config(
+    $foreman_user  = 'foreman',
+    $foreman_group = 'foreman',
+    $foreman_plugins_dir = '/etc/foreman/plugins',
+    $protocol = 'https',
```
ekohl: I still think a single URL parameter is easier than all the separate components. Now you're still forced to use the fqdn which a user might not want.

dustints:
@ekohl updated to a single url param

manifests/plugin/gutterball.pp
ekohl: I still think you should move the file resource into this class and drop the separate ::config class because it will simplify things.

manifests/plugin/gutterball/config.pp
```diff
@@ -0,0 +1,15 @@
+# gutterball config
+class katello::plugin::gutterball::config(
+    $foreman_user  = 'foreman',
+    $foreman_group = 'foreman',
+    $foreman_plugins_dir = '/etc/foreman/plugins',
+    $url = "https://${fqdn}:8443/gutterball",
+  ){
+
+  file { "${foreman_plugins_dir}/foreman_gutterball.yaml":
```
ekohl: I'd name this simply gutterball.yaml since foreman is implied.

dustints:
@ekohl, updated renamed the soruce template & destination config file

manifests/plugin/gutterball/config.pp
ehelms: I prefer things are explicit and match the plugin name. That way there is
no translation or discrepancy whenever you are looking for something
related. I think that the namespace of the plugins config and the file
containing it should remain the same.

ekohl: That's also a good point. I'll think about this and get back to you.

komidore64: I agree with @ehelms. the name of the plugin is `foreman_gutterball` so the plugin's config file should have the same name.

dustints:
merging! too much time spent on this 10 liner

PR #59, opened 2015-02-06
refs #9060 - configure qpid::client with params

stbenjam:
Add parameters for configuring the client, needed for the changes that happened in https://github.com/Katello/puppet-qpid/pull/12

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -97,7 +97,12 @@ class katello (
     key     => $certs::apache::apache_key,
     ca_cert => $certs::ca_cert,
   } ~>
-  class { 'qpid::client': } ~>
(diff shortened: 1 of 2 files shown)


ehelms:
@stbenjam what about a basic test to ensure this client is called with this config?

stbenjam:
Here's a test.

I think the test is of dubious value, I always feel wrong pulling variables across namespaces (e.g. pulling certs values from katello), but I'm not enough of a puppet-y person to know if my instinct is right...  and I'm not sure how to get at the values in rspec to test for all the parameters. Any idea?

Regardlesss, the habit of writing/asking for tests is at least good :+1:

ehelms:
ACK

PR #61, opened 2015-02-19
Fixes #7716: Restart foreman-tasks when katello.yml changes.

ehelms:


jlsherrill:
Should this also notify the httpd service?

ehelms:
Good call. I was just blindly following the bug.

ehelms:
@jlsherrill Added httpd

manifests/config.pp
```diff
@@ -15,6 +15,7 @@
     group   => $katello::group,
     mode    => '0644',
     before  => [Class['foreman::database'], Exec['foreman-rake-db:migrate']],
+    notify  => [Service['foreman-tasks'], Service['httpd']],
```
ekohl: I thought this was the same as https://github.com/Katello/puppet-katello/pull/62, but also here you may want to notify ```Class['foreman::service']```.

ehelms:
Green here to!

iNecas:
ACK

PR #68, opened 2015-04-16
refs #6781 - add services for katello to services.d

stbenjam:


manifests/config.pp
```diff
@@ -30,4 +30,6 @@
     mode   => '0755',
   }
 
+  $services_real = prefix($katello::services_list, $katello::services_directory)
+  ensure_resource('file', $services_real, {'ensure' => 'file', 'mode' => '0644'})
```
ekohl: Any reason you're not simply using a file block?
```puppet
file { $services_real:
    ensure => 'file',
    mode   => '0644',
}
```

manifests/params.pp
```diff
@@ -63,4 +63,15 @@
...
   $candlepin_event_queue = 'katello_event_queue'
+
+
+  if (versioncmp($::operatingsystemrelease, '7.0') < 0) {
+    $tomcat_service = 'tomcat6'
+  } else {
+    $tomcat_service = 'tomcat'
+  }
+
+  $services_directory = '/etc/katello/services.d'
```
ekohl: Shouldn't this include a trailing slash since you're simplify prefixing in the code?

manifests/init.pp
```diff
@@ -58,8 +63,12 @@
   $cdn_ssl_version = $katello::params::cdn_ssl_version,
 
   $package_names = $katello::params::package_names,
+
+  $services_directory = $katello::params::services_directory,
+  $services_list      = $katello::params::services_list,
   ) inherits katello::params {
 
+
```
ekohl: Nitpick: why this additional line?

manifests/init.pp
```diff
@@ -35,6 +35,11 @@
 #
 # $package_names::      Packages that this module ensures are present instead of the default: katello and ${scl_prefix}rubygem-katello
 #
+# $services_directory:: Directory where services configuration is stored
```
ekohl: Can be validated with ```validate_absolute_path``` from stdlib.

manifests/init.pp
```diff
@@ -35,6 +35,11 @@
...
+#
+# $services_list::      A list of services for the katello-service tool
```
ekohl: Can be validated with ```validate_array``` from stdlib.

manifests/config.pp
stbenjam: It could be that puppet-katello and puppet-capsule end up with overlapping services.

ekohl: In that case ensure_resource is dependant on the order (unless both use ensure_resource). Maybe a step back: why are these files needed even though they don't have any content (here)? In case you do want content, wouldn't a define be better? Similar to how we have ```foreman::plugin``` and then many ```foreman::plugin::*``` classes.

stbenjam: They're both `ensure_resource`.

 `katello-service` tool will look services.d and restart all the services by the filename.  It's simple, but your comment makes me realize this approach doesn't work.

If you create say, `/etc/katello/services.d/pulp` on a capsule and later remove the service, the directory's not going to get cleared (and can't, it's managed by two modules).

I'll find a different approach

stbenjam: Thanks @ekohl

ehelms:
Do we still need this?

stbenjam:
Nope

PR #75, opened 2015-08-11
fixes #11326 - fixes pulp isos from being inaccessible via httpd

cfouant:


stbenjam:
Seems reasonable to me ACK

ehelms:
Thanks @cfouant !

PR #77, opened 2015-08-28
Fixes #11609: Update to use Pulp module 1.0

ehelms:


manifests/init.pp
```diff
@@ -83,18 +80,13 @@
...
-    qpid_ssl_cert_password_file => $certs::qpid::nss_db_password_file,
     messaging_ca_cert           => $certs::pulp_parent::messaging_ca_cert,
     messaging_client_cert       => $certs::pulp_parent::messaging_client_cert,
-    consumers_ca_cert           => $certs::ca_cert,
-    consumers_ca_key            => $certs::ca_key,
     consumers_crl               => $candlepin::crl_file,
     proxy_url                   => $proxy_url,
     proxy_port                  => $proxy_port,
     proxy_username              => $proxy_username,
     proxy_password              => $proxy_password,
```
stbenjam: The `=>`'s should be aligned one space to the right of the longest parameter

stbenjam:
Is 1.0 ready?

ehelms:
I believe so functional parity wise - I wanted to do this part and do a test round with the installer, however, that is hard to do right now in the middle of TFM migration. This PR does remove an input parameter from this module.

ehelms:
@stbenjam pushed updates to what we need to pass to Pulp and deploy, my local testing says this should be good to go along with -- https://github.com/Katello/puppet-pulp/pull/97

stbenjam:
ACK

PR #82, opened 2015-09-11
Set the adapter module to use in Candlepin

ehelms:


ehelms:
Closed in favor of including here -- https://github.com/Katello/puppet-katello/pull/83

PR #84, opened 2015-09-13
Don't pass ca cert/key to Pulp

ehelms:
The ca_key and ca_cert options are used by Pulp when creating and
signing consumer certs. Not only do we not use this ability of Pulp
but these options are being deprecated by Pulp.


jlsherrill:
ACK

PR #88, opened 2015-10-08
Fixes #12110 - updating puppet-pulp to use is_parent => true param

johnpmitsch:


manifests/params.pp
```diff
@@ -64,4 +64,6 @@
 
   $qpid_url = "amqp:ssl:${::fqdn}:5671"
   $candlepin_event_queue = 'katello_event_queue'
+
+  $is_parent = true
```
stbenjam: Is this intended to be used somewhere?

johnpmitsch: trying to update for this change https://github.com/Katello/puppet-pulp/pull/108

stbenjam: params.pp is only if the parameter is actually a class parameter (in this case on katello  itself).

manifests/init.pp
```diff
@@ -109,6 +109,7 @@
     enable_puppet         => true,
     enable_docker         => true,
     num_workers           => $num_pulp_workers,
+    is_parent             => true,
```
johnpmitsch: @stbenjam so basically I would just need this?

johnpmitsch:
Waiting on https://github.com/Katello/puppet-pulp/pull/108 to be merged

manifests/init.pp
stbenjam: Yup, that's it. Only need the params.pp in the case you have in the other PR, when adding it to the class's parameters list.

stbenjam:
Looks good to me,  APT

ehelms:
Thanks @johnpmitsch ! Please do the same to puppet-katello_devel

PR #89, opened 2015-10-08
Use cache_data and random_password from extlib

ehelms:


stbenjam:
I think it's going to cause issues that foreman's method is named the same, will they move to extlib?

ehelms:
See https://github.com/theforeman/puppet-foreman/pull/364

stbenjam:
Great :-)

We'll need puppet-foreman/foreman_proxy to update their versions so we can set the right requirements in metadata.json as well

stbenjam:
APT, should go green when https://github.com/Katello/puppet-certs/pull/68 is merged

PR #90, opened 2015-10-26
Fixes #12309: Deploy Candlepin on to port 9073

ehelms:
Currently base Foreman proxies deploy on 8443, and Katello capsules
deploy on 9090 (i.e. deploy Foreman proxies on 9090) because of
Candlepin being deployed on 8443. Since Candlepin is used on for
internal access, moving it to a different port helps resolve this
inconsistency.


ehelms:
Requires -- https://github.com/Katello/puppet-candlepin/pull/29

stbenjam:
This will require SElinux changes too, to remove the old proxy port and add the new candlepin port.

Unless we pick a port for candlepin that's already allowed for http:

```
[root@sat-perf-01 ~]# semanage port -l | grep -w http_port_t
http_port_t                    tcp      80, 81, 443, 488, 8008, 8009, 8443, 9000
```

ehelms:
Good catch -- what started this is the fact that we have three services that by default want to use 8443:

smart proxy
candlepin
reverse proxy on the Capsule

Changing the internally referenced Candlepin seemed like the easiest place to start. This plays into the whole idea of having an "infrastructure" port deployed for internal communication.

stbenjam:
It's a good idea - I was just suggesting maybe using 9000 for candlepin instead, so any selinux related rules that apply to the current port (8443, which is http_port_t) apply to the new one too.

For the RHSM proxy on the Capsule, would it be possible to move that to run on 443, or is it because of the certs stuff that it can't? I think there's a downstream bugzilla for that even, or at least a downstream user was asking about it.

Then we can keep the foreman default 8443 for smart proxy everywhere.

Hopefully we try to get this all in one release (maybe 2.5) so we're not making people change their firewall rules more than once.

ehelms:
The problem comes with all the various services we can deploy and their ports, currently that is:

Katello
web server : 443
smart proxy : 9090
pulp : 443
candlepin : 8443
reverse proxy on capsule : 8443
smart proxy capsule : 9090
capsule content + gpg keys: 443

Foreman
web server : 443
smart proxy : 8443

Now if we introduce infrastructure port concept that needs to live at a particular port as well. If we pick pre-configured ports, then we are limiting ourselves and risking clashing with other services. Versus if we picked some unused port range that we deployed all of our services to we have better control and visibility (IMO). For example (keeping in mind this is greenfield):

443 - Web Server
9070 - Infrastructure Port (would include reverse proxy on Capsules)
9071 - Smart Proxy
9072 - Pulp
9073 - Candlepin

stbenjam:
Ok then, that works for me. APT

ekohl:
Requires a rebase.

ekohl:
Any plan to revisit this?

ekohl:
Closing for inactivity.

PR #91, opened 2015-10-29
[messaging] disable authentication

bbuckingham:
With the puppet-pulp 1.0 module, the default for auth_enabled on messaging
was to enable it.  This breaks our current configuration; therefore, this
change is to set the value back to what we had prior.

Note: Enabling this feature will enable message signing and verification.
Since katello agent does not sign messages, they are discarded
by the server.


ehelms:
ACK

PR #92, opened 2015-11-17
Fixes #12475: Configure Candlepin AMQP since it is off by default

ehelms:


stbenjam:
Needs rebasing

stbenjam:
Can you bump the minor here, and the requirement for puppet-candlepin to `>= 0.2.0`? Since some of these params are new

ehelms:
Updated

metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "katello-katello",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
stbenjam: The patch level was already pre-bumped, it works like we do with tito (2.0.1 is the version of the forge).

So maybe we can leave it as 2.0.2, I guess I was wondering what the semver way to handle changing dependencies, any thoughts there? I guess since the signature of puppet-katello didn't change patch only is fine.

thomasmckay:
ACK from testing puppet-katello_devel PR

metadata.json
ehelms: Yea - we aren't changing an external "API" with this. I put this back to 2.0.2

stbenjam:
Looks good, ACK from me!

ekohl:
Thanks!

PR #94, opened 2016-01-15
add a truststore_password parameter for candlepin

cristifalcas:
This adds support for truststore password for candlepin, as pushed in PR#32.

I will also open a PR for puppet-certs to add the corresponding parameter

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -76,7 +76,7 @@ class katello (
     ca_key                       => $certs::ca_key,
     ca_cert                      => $certs::ca_cert_stripped,
     keystore_password            => $::certs::candlepin::keystore_password,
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -76,6 +76,7 @@
     ca_key                       => $certs::ca_key,
     ca_cert                      => $certs::ca_cert_stripped,
     keystore_password            => $::certs::candlepin::keystore_password,
+    truststore_password          => $::certs::candlepin::truststore_password,
```
ehelms: You can just set this to use the keystore password from certs

cristifalcas:
Should I close the other PR from certs now?

ehelms:
ACK - thanks @cristifalcas !

PR #95, opened 2016-01-19
Fixes #13189 - allows for certificate authentication of capsules

cfouant:
Requires https://github.com/Katello/puppet-certs/pull/70

### DIFF
--- a/templates/etc/httpd/conf.d/05-foreman-ssl.d/katello.conf.erb
+++ b/templates/etc/httpd/conf.d/05-foreman-ssl.d/katello.conf.erb
@@ -3,7 +3,7 @@
 # CHANGES WILL LIKELY BE OVERWRITTEN.
 #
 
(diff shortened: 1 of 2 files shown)


templates/etc/httpd/conf.d/05-foreman-ssl.d/katello.conf.erb
```diff
@@ -3,7 +3,7 @@
 # CHANGES WILL LIKELY BE OVERWRITTEN.
 #
 
-SSLOptions +StdEnvVars +ExportCertData +FakeBasicAuth
+SSLOptions +StdEnvVars +ExportCertData
```
ekohl: I think you can drop this since it now matches the default from puppet-foreman.

jlsherrill:
ACK, thanks @cfouant

PR #101, opened 2016-02-10
Fixes #13503 - Updating post sync url

parthaa:
Updating server post sync url to fqdn since it needs to match the cn in
the katello cert and not (localhost)


ekohl:
ACK

ehelms:
Thanks @parthaa !

PR #106, opened 2016-02-13
Refs #13658 - remove cycle from puppet graph

beav:
The previous commit (https://github.com/Katello/puppet-katello/commit/4ecc339d16b88f1a450f075fc14e0c0a103a3e4a) added two cycles to the puppet dep graph, causing `katello-installer` to fail.

The offending line was:

```
require        => [Exec['foreman-rake-db:seed'], Class['::certs::pulp_client']],
```

Note that each of these `require`s individually will cause a cycle. The cycle for just `foreman-rake-db:seed` is http://i.imgur.com/eA8rWc6.png and the cycle for `::certs::pulp_client` is http://i.imgur.com/EgeI2X2.png.

Removal of these lines makes `katello-installer` succeed.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -3,6 +3,8 @@ class katello::config {
 
   $apache_version = $::apache::apache_version
 
+  class { '::katello::config::pulp_client': }
+
   file { '/usr/share/foreman/bundler.d/katello.rb':
     ensure => file,
     owner  => $katello::user,
@@ -32,16 +34,5 @@ class katello::config {
     mode   => '0755',
   }
 
-  foreman_config_entry { 'pulp_client_cert':
-    value          => $::certs::pulp_client::client_cert,
-    ignore_missing => false,
-    require        => [Exec['foreman-rake-db:seed'], Class['::certs::pulp_client']],
-  }
-
-  foreman_config_entry { 'pulp_client_key':
-    value          => $::certs::pulp_client::client_key,
-    ignore_missing => false,
-    require        => [Exec['foreman-rake-db:seed'], Class['::certs::pulp_client']],
-  }
 
 }
--- a/manifests/config/pulp_client.pp
+++ b/manifests/config/pulp_client.pp
@@ -1,0 +1,18 @@
+#Katello Pulp Client config
+class katello::config::pulp_client {
+
+  class { '::certs::pulp_client': }
+
+  foreman_config_entry { 'pulp_client_cert':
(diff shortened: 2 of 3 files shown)


ehelms:
However, this removes the guarantee that this is run at the correct time given this requires the certs exist and db is migrated. The better fix might be to remove the pulp client certs from the giant dependency block in init.pp.

beav:
@ehelms katello::config was needed to seed the db, but the db seeding was needed for katello::config. I broke out the pulp client config so it can happen after the db seeding.

Additionally I implemented your other suggestion for the other cycle.

:no_bicycles: :no_bicycles: :no_bicycles:

manifests/config.pp
```diff
@@ -32,16 +32,7 @@
...
-    ignore_missing => false,
-    require        => [Exec['foreman-rake-db:seed'], Class['::certs::pulp_client']],
-  }
-
-  foreman_config_entry { 'pulp_client_key':
-    value          => $::certs::pulp_client::client_key,
-    ignore_missing => false,
-    require        => [Exec['foreman-rake-db:seed'], Class['::certs::pulp_client']],
-  }
+  class { '::certs::pulp_client': }
```
ehelms: Might as well move this into the pulp_client_config class, and I might consider just using the namespace, e.g.

```
config/pulp_client.pp
```

```
katello::config::pulp_client
```

beav:
latest commit has a cycle, fixing

beav:
@ehelms I lied, sorry. Current PR passes my test and is ready for review.

ehelms:
I think this looks good, @ekohl mind weighing in on the design change here?

ehelms:
I will porbably need to test this in the installer, the current implementation results in:

```
[ WARN 2016-02-15 10:32:03 verbose]  Scope(Class[Katello::Config]): Could not look up qualified variable '::certs::pulp_client::client_cert'; class ::certs::pulp_client has not been evaluated
[ WARN 2016-02-15 10:32:03 verbose]  Scope(Class[Katello::Config]): Could not look up qualified variable '::certs::pulp_client::client_key'; class ::certs::pulp_client has not been evaluated
```

ekohl:
I can't judge if the ```pulp_client``` ~> ```pulp_parent``` relation is needed (and in turn pulp and qpid), but it sounds like it shouldn't be.

ehelms:
This passed my installer testing - thanks @beav !

PR #108, opened 2016-02-17
Fixes #13625 - Install Ostree packages on enablement

parthaa:


spec/classes/katello_install_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it { expect contain_package("tfm-rubygem-katello-ostree").with_ensure('installed') }
+  end
+
+  context 'should NOT install ostree package on enable_ostree == false' do
+    let(:pre_condition) do
+      [
+        'class {"katello":' \
+          'enable_ostree => false,' \
+        '}'
+      ]
```
ehelms: I would do this like https://github.com/Katello/puppet-pulp/blob/master/spec/classes/pulp_config_spec.rb#L62

manifests/install.pp
```diff
@@ -3,4 +3,10 @@
   package { $katello::package_names:
     ensure => installed,
   }
+
+  if ($katello::enable_ostree) {
```
ehelms: I don't think the parens are necessary here or typical.

manifests/install.pp
```diff
@@ -3,4 +3,10 @@
...
+    package { [$katello::params::rubygem_katello_ostree]:
```
ehelms: Don't reference params outside of the class parameter definitions -- also I don't see 'rubygem_katello_ostree' defined anywhere.

parthaa: referenced here -> https://github.com/Katello/puppet-katello/pull/108/files#diff-2f383a197b1a428561b364755701d60eR12

manifests/init.pp
```diff
@@ -37,6 +37,9 @@
 #
 # $package_names::      Packages that this module ensures are present instead of the default
 #
+# $enable_ostree::      Boolean to enable ostree plugin. This requires existence of a ostree install.Defaults to false
```
ekohl: In the other PR we removed the defaults to part.

manifests/install.pp
```diff
@@ -3,4 +3,10 @@
   package { $katello::package_names:
     ensure => installed,
   }
+
+  if $katello::enable_ostree {
+    package { [$katello::params::rubygem_katello_ostree]:
```
ekohl: I'd remove the array part. You can also refer to ```$katello::rubygem_katello_ostree``` since ```katello``` inherits from ```katello::params```.

manifests/params.pp
```diff
@@ -9,6 +9,7 @@
         }
         default: {
           $rubygem_katello = 'tfm-rubygem-katello'
+          $rubygem_katello_ostree ='tfm-rubygem-katello-ostree'
```
ekohl: It should also be defined for other operating systems. In the foreman modules we always set it even if it's unpackaged. That way you can package it in a minor/patch release without changing the modules.

spec/classes/katello_install_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      [
+        "class {'katello':
+          enable_ostree => true,
+        }"
+      ]
+    end
+    it { expect contain_package("tfm-rubygem-katello-ostree").with_ensure('installed') }
+  end
+
+  context 'should NOT install ostree package on enable_ostree == false' do
```
ekohl: The context is with ostree disabled. Expectations are not part of the context.

manifests/params.pp
parthaa: @ekohl updated.

manifests/init.pp
parthaa: I think the defaults make sense here though. Because we have other modules calling this to pass parameters, its useful to know what the default is.

ekohl: I don't mind, but in that case I'd at least place a space between the dot and Defaults for readability.

ekohl:
:+1: besides the nitpick.

ehelms:
ACK - thanks @parthaa !

PR #109, opened 2016-02-18
fixes #13451 - enables squid management in pulp

cfouant:


daviddavis:
I think we also want to set `$lazy_download_interval` to something less than 30 minutes.

cfouant:
@daviddavis - okay, that setting is in puppet-pulp though. I'll adjust.

daviddavis:
Some things I noticed when doing a nightly install:

1. In server.conf, redirect_port was set to `undef` 
2. The python-pulp-streamer package wasn't installed. Just is rebuilding the katello rpm which should include it in the nightlies but I think we'll need to update these files to install the package for development environments:
  - https://github.com/Katello/puppet-pulp/blob/master/manifests/install.pp
  - https://github.com/Katello/puppet-pulp/blob/master/manifests/service.pp

daviddavis:
A couple things I found while debugging:

- In the version of pulp we're using in the nightlies, we need to set "enabled: true" in the lazy sync section of pulp's server.conf
- The /etc/httpd/conf.d/pulp_streamer.conf is missing. I believe puppet wipes this out.
- Also there seems to be some funkiness around SSL. Investigating further.

jlsherrill:
> In the version of pulp we're using in the nightlies, we need to set "enabled: true" in the lazy sync section of pulp's server.conf

that is true for the current katello nightly, but via the latest pulp nightlies and the next katello nightly (whenever it passes the pipeline) its not needed.  So i'd say leave it out for now and by the time this hits nightly, the new pulp will be there

daviddavis:
ACK from me.

stbenjam:
As you're including a new parameter, the requirement in `metadata.json` needs to be updated, and the patch (or whatever's appropriate) level bumped in the corresponding class.

Otherwise you could end up getting katello-pulp 3.0.0, which does not support manage_squid.

cfouant:
@stbenjam - I believe they will be added once https://github.com/Katello/puppet-pulp/pull/130 merges.

stbenjam:
Where? manage_squid was already added, katello-pulp version needs to be bumped and released.  Likewise, the requirement in this PR in the metadata.json here https://github.com/Katello/puppet-katello/blob/master/metadata.json#L57 needs to be updated.

It's why tests are failing.

stbenjam:
Although it should be getting puppet-pulp in the fixtures from git for the tests, so not sure why tests are failing actually :confused:

ehelms:
I released puppet-pulp 3.1.0 so the metadata can be updated here

stbenjam:
Oh I see, [test]s were run before it was added. I'd guess it's green now pulling from git.  Re: the version requirement though, it's something that still probably should be done to make sure we try to be as accurate as possible in the dependencies.

daviddavis:
Is this good? https://github.com/Katello/puppet-katello/pull/112

stbenjam:
Sure but imho it should be combined with the PR that introduces the new requirement, makes it easier to look later to see what it was that required 3.1.0.

daviddavis:
Updated https://github.com/Katello/puppet-katello/pull/112

ehelms:
Closing due to https://github.com/Katello/puppet-katello/pull/112

PR #110, opened 2016-02-18
Fixes #13605 - Configure keep alive for 443 virtual host

jlsherrill:


jlsherrill:
I made this configurable based on the pulp's recommendation:  https://github.com/pulp/pulp/blob/master/server/etc/httpd/conf.d/pulp_apache_24.conf#L67-L79

jlsherrill:
also going to add capsule support

manifests/init.pp
```diff
@@ -40,6 +40,9 @@
 # $enable_ostree::      Boolean to enable ostree plugin. This requires existence of an ostree install.
 #                       type:boolean
 #
+# $max_keep_alive::     Maximum number of requests to use for the apache MaxKeepAliveRequests apache paramter
```
ehelms: I think you can drop the second apache. ALso, a typo in "parameter"

templates/etc/httpd/conf.d/05-foreman-ssl.d/katello.conf.erb
```diff
@@ -26,3 +26,6 @@ Alias /pub /var/www/html/pub
   ErrorDocument 500 '{"displayMessage": "Internal error, contact administrator", "errors": ["Internal error, contact administrator"], "status": "500" }'
   ErrorDocument 503 '{"displayMessage": "Service unavailable or restarting, try later", "errors": ["Service unavailable or restarting, try later"], "status": "503" }'
 </LocationMatch>
+
+KeepAlive On
+MaxKeepAliveRequests <%= @max_keep_alive %>
```
ehelms: Prefer `scope['::katello::max_keep_alive']` syntax

jlsherrill:
@ehelms updated

manifests/init.pp
```diff
@@ -64,6 +67,7 @@
 
   $package_names = $katello::params::package_names,
   $enable_ostree = $katello::params::enable_ostree,
+  $max_keep_alive = $katello::params::max_keep_alive,
```
ekohl: Can be validated using ```validate_integer``` (available in stdlib >= 4.6.0)

ekohl: Btw, not sure what kafo does or if you need type:integer as well in the documentation.

jlsherrill: The foreman module seems to use it, but i don't see it used much in the katello module.  Went ahead and added it.

ekohl: You need to increase the minimum version in metadata.json as well.

beav:
This works as advertised when tested, ACK from me.

ehelms:
Once https://github.com/Katello/puppet-katello/pull/110#discussion_r53462482 is addressed ACK

PR #111, opened 2016-02-23
ignore MaxKeepAliveRequests if not set

jlsherrill:


jlsherrill:
The reason for this is that this template is included here:  https://github.com/Katello/puppet-katello_devel/blob/master/manifests/apache.pp#L25

This isn't a great solution, but i'm not positive what the proper solution is (other than keeping two copies of the config  in each module) which i don't really like

ehelms:
I think this is a fair and proper solution. ACK for me

beav:
devel installer passes, ACK
:panda_face:

PR #115, opened 2016-03-08
Fixes 13199 - remove pulp-nodes packages as dependency

johnpmitsch:


ehelms:
ACK - thanks @johnpmitsch !

PR #119, opened 2016-03-28
Simplify variables in katello::qpid

ekohl:


manifests/qpid.pp
```diff
@@ -1,8 +1,9 @@
-# Katello Config
+# Katello qpid Config
 class katello::qpid (
-  $client_cert            = undef,
-  $client_key             = undef,
-  $katello_user           = $katello::user
+  $client_cert,
+  $client_key,
+  $katello_user          = $::katello::user,
+  $candlepin_event_queue = $::katello::candlepin_event_queue,
```
stbenjam: I don't see this variable in `$::katello`, the exec below was referencing `$::katello::params::candlepin_event_queue`  Maybe candlepin_event_queue could be a param on the qpid class itself?

stbenjam:
Otherwise, looks good - and thanks, nice cleanup

manifests/qpid.pp
ekohl: Right now the katello class inherits from params so that way it should be available. By using it this way you can easily make it a real parameter later on.

ehelms:
@stbenjam I believe this is waiting on you

stbenjam:
ACK, thanks @ekohl

PR #120, opened 2016-04-12
Revert "Simplify variables in katello::qpid"

stbenjam:
Reverts Katello/puppet-katello#119

This broke the pipeline:

```
21:49 [-beav > ==> centos7-devel:  qpid-config --ssl-certificate /etc/pki/katello/certs/centos7-devel.example.com-qpid-broker.crt --ssl-key                                                                        
               /etc/pki/katello/private/centos7-devel.example.com-qpid-broker.key -b 'amqps://centos7-devel.example.com:5671' bind event  '*.*' returned 1 instead of one of [0] 
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -139,8 +139,9 @@ class katello (
     ssl_cert_password_file => $certs::qpid::nss_db_password_file,
   } ~>
   class { '::katello::qpid':
-    client_cert => $certs::qpid::client_cert,
-    client_key  => $certs::qpid::client_key,
+    client_cert  => $certs::qpid::client_cert,
+    client_key   => $certs::qpid::client_key,
+    katello_user => $katello::user,
   } ~>
   Exec['foreman-rake-db:seed']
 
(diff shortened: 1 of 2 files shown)


stbenjam:
For some reason `$candlepin_event_queue` doesn't seem to be in scope.  Maybe @ekhol has an idea why, but I'd be inclined to just revert this for now and revisit the clean-up later so the pipeline gets going again.

beav:
the pipeline itself is green, this is for devel setup but is also important :koala: 

I'll do a run w/ this patch asap to see if it runs cleanly

beav:
I get a clean run after applying this patch and re-running foreman-installer (I was already set up to use the devel scenario).

stbenjam:
Ohhh sorry, thought I saw somethng on foreman-dev earlier.  This is still likely the cause.  For some reason `$candlepin_event_queue` isn't available when puppet-katello_devel calls `::katello::qpid`.

stbenjam:
I see, katello_devel calls `::katello::qpid`, but the top-level namespace (`::katello`) where candlepin_event_queue is, isn't loaded.

I guess the best fix would be for katello_devel to pass the queue explicitly.

This should fix it: https://github.com/Katello/puppet-katello_devel/pull/72

Will test.

But no need to revert this change IMHO.

PR #121, opened 2016-04-13
Fixes #14617 - communicate with candlepin over fqdn

jlsherrill:


jlsherrill:
doesn't seem like a katello_devel update is needed

beav:
hooray for cert validation :fireworks: 

ACK

PR #122, opened 2016-04-16
Add paths for puppet-lint docs check

stbenjam:
This checks that each of the class parameters is documented. We need this check
on modules exposed in the installer directly.


ekohl:
ACK

stbenjam:
Thanks!

PR #123, opened 2016-04-18
Refs #14698 - create default directory for repo exports

beav:
Previously, there was no default location for repository exports. This meant
that users had to "leave" Katello and hop into the command line prior to
creating their first repo export.

This commit creates a `/var/lib/pulp/katello-export` directory as part of the
install. Users may optionally point this elsewhere via installer parameter if
needed.

The `/var/lib/pulp/katello-export` directory was chosen since users already
expect that `/var/lib/pulp` and its child dirs may have large disk space
requirements.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -34,5 +34,20 @@ class katello::config {
     mode   => '0755',
   }
 
+  # NB: we define this here to avoid a dependency cycle. It is not a problem if
+  # this dir exists before the pulp RPMs are installed.
+  file { '/var/lib/pulp':
+    ensure => directory,
+    owner  => 'apache',
+    group  => 'apache',
+    mode   => '0755',
+  }
+
+  file { $katello::repo_export_dir:
+    ensure => directory,
+    owner  => $katello::user,
+    group  => $katello::group,
+    mode   => '0755',
+  }
 
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -44,6 +44,8 @@
 #                       on the virtualHost for port 443.
 #                       type: integer
 #
+# $repo_export_dir::    Directory to create for repository exports
+#
 class katello (
 
   $user = $katello::params::user,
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -69,6 +71,8 @@
   $package_names = $katello::params::package_names,
   $enable_ostree = $katello::params::enable_ostree,
   $max_keep_alive = $katello::params::max_keep_alive,
+
+  $repo_export_dir = $katello::params::repo_export_dir,
```
ekohl: This can be verified using ```validate_absolute_path```.

ekohl:
ACK

manifests/init.pp
beav: that is neat! I was not aware of `validate_absolute_path`

beav:
patch updated to validate path

manifests/init.pp
```diff
@@ -148,5 +153,16 @@
 
   Service['httpd'] -> Exec['foreman-rake-db:seed']
 
+  file { $repo_export_dir:
```
ehelms: This seems more appropriate in the config manifest based on our typical separation of duties

beav: that makes more sense, I will update

beav:
@ehelms updated per comments

manifests/config.pp
```diff
@@ -34,5 +34,15 @@
...
 
+  file { $katello::repo_export_dir:
+    ensure  => directory,
+    owner   => $katello::user,
+    group   => $katello::group,
+    mode    => '0755',
+    # the default export dir is /var/lib/pulp/katello-export, this 'require'
+    # ensures that the parent dirs exist without having to define the full
+    # tree.
+    require => Package['pulp-server'],
```
ehelms: Does this error if pulp-server isn't installed yet?

beav: It shouldn't? I don't see why it would error for a requires, it should just install the pkg if its not there.

beav:
@ehelms  updated per in-person comments last week

manifests/config.pp
```diff
@@ -34,5 +34,22 @@
...
+    mode    => '0755',
+  }
+
+  file { $katello::repo_export_dir:
+    ensure  => directory,
+    owner   => $katello::user,
+    group   => $katello::group,
+    mode    => '0755',
+    # the default location requires this parent dir to exist
+    require => File['/var/lib/pulp']
```
ekohl: Since puppet 2.7 file resources auto-require all parents if they exist as puppet resources, so this should be redundant.

ehelms:
Thanks @beav !

PR #125, opened 2016-05-10
Remove 'tomcat' param

kmcfate:
Tomcat not used and 'file_exists' is master side only. This does not work as expected except when using the masterless installer.


ekohl:
ACK. looks like it was introduced in ae7d4aa1039445bbf130d9951ab36d866d551a3b but I can't see it used anywhere.

stbenjam:
Yup, ACK from me too - I didn't see any usage of this anywhere

ekohl:
Thanks!

PR #126, opened 2016-05-16
fixes #15058 - enable mongo auth

stbenjam:
Companion PR's:
- [ ] https://github.com/Katello/puppet-pulp/pull/146
- [ ] https://github.com/Katello/puppet-capsule/pull/84
- [ ] https://github.com/Katello/puppet-katello_devel/pull/75

Test box:
```
mongoauth:
    box: centos7-katello-nightly
    options: --module-prs katello/126,pulp/146,capsule/84
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -46,6 +46,10 @@
 #
 # $repo_export_dir::    Directory to create for repository exports
 #
+# $pulp_db_username::   Username for the pulp database
+#
+# $pulp_db_password::   Password for the pulp database
+#
 class katello (
 
   $user = $katello::params::user,
@@ -73,6 +77,9 @@ class katello (
   $max_keep_alive = $katello::params::max_keep_alive,
 
   $repo_export_dir = $katello::params::repo_export_dir,
+
+  $pulp_db_username = $katello::params::pulp_db_username,
+  $pulp_db_password = $katello::params::pulp_db_password,
   ) inherits katello::params {
   validate_bool($enable_ostree)
   validate_integer($max_keep_alive)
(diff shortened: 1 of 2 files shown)


ekohl:
ACK

manifests/init.pp
```diff
@@ -46,6 +46,10 @@
 #
 # $repo_export_dir::    Directory to create for repository exports
 #
+# $pulp_db_username::   Username for mongo
+#
+# $pulp_db_password::   Password for mongo
```
ehelms: We aren't doing this kind of exposure for the Candlepin database, maybe we should. Also, if you change this to just database we won't have to update it when Pulp switches off mongo.

stbenjam:
@ehelms Renamed mongo to database in the docs

stbenjam:
On the other PR's, too

ekohl:
ACK

beav:
I don't think this works on el6, there is a new el6 bats failure:

```

# [[31mERROR[0m [34m2016-05-17 17:55:26[0m [36mverbose[0m]  Could not prefetch mongodb_user provider 'mongodb': Execution of '/usr/bin/mongo admin --quiet --host 127.0.0.1:27017 --eval printjson(db.isMaster())' returned 1: Tue May 17 17:55:26.480 Error: couldn't connect to server 127.0.0.1:27017 at src/mongo/shell/mongo.js:147

````

akofink:
This fails on CentOS7 (centos7-devel box):

`/usr/bin/mongo admin --quiet --host 127.0.0.1:27017 --eval printjson(db.isMaster())`

This works, however (with quotes):

`/usr/bin/mongo admin --quiet --host 127.0.0.1:27017 --eval 'printjson(db.isMaster())'`

I'm just executing these commands on the centos7-devel box after running `vagrant up`, which is currently failing at these pulp_db setup commands.

stbenjam:
The quotes would probably be removed when it shows you the command but they should be there when puppet's actually executing the command. I think.  It's quirky.

I've seen this error intermittently, and @beav saw some other Mongo-related issues.  I wonder if we just end up turning this back off all together. It only works on one OS, and apparently not very well.

Mongo is going away eventually anyway.

PR #129, opened 2016-05-24
Fixes #15170 - Moving timeout to a large enough value

mccun934:
This is necessary because we have various operations in Katello that
are synchronus, especially around our interaction with candlepin and
manifest import that we will often time-out during operation.

We need to keep this timeout value high until the API calls we rely on
are implemented in an async manner.


bbuckingham:
ACK

lzap:
Can you guys check this gets really honored? I tested this patch and it does not change a thing: https://bugzilla.redhat.com/show_bug.cgi?id=1334996#c5

PR #130, opened 2016-06-07
refs #15326 - revert mongo auth

stbenjam:


akofink:
So we'll need to release tag 5.0.0 with this PR, right?

ACK

ehelms:
ACK

PR #135, opened 2016-07-18
Fixes #15727 - Not receiving candlepin messages

johnpmitsch:
Messages are not being received from candlepin due to exec
blocks not being executed sequentially. The 2nd exec{} is dependent
on the 1st exec{} being completed successfully.


johnpmitsch:
Tested this out with a fresh install and can confirm that candlepin messages are being sent with this change.

jlsherrill:
ACK

PR #136, opened 2016-07-26
Fixes #15841 - limit pulp puppet wsgi procs

jlsherrill:
This is a little used feature and extra wsgi
processes use up more open file handles


ekohl:
Depends on https://github.com/Katello/puppet-pulp/pull/155 but ACK on green tests

ehelms:
Tests are green - merging, thanks @jlsherrill !

PR #137, opened 2016-08-03
Wrap `PassengerEnabled` in module check

beav:
Previously, this vhost def was disabling passenger even if passenger was not
installed, causing errors.

### DIFF
--- a/templates/etc/httpd/conf.d/05-foreman.d/katello.conf.erb
+++ b/templates/etc/httpd/conf.d/05-foreman.d/katello.conf.erb
@@ -6,7 +6,9 @@
 Alias /pub /var/www/html/pub
 
 <Location /pub>
-  PassengerEnabled off 
+  <IfModule mod_passenger.c>
+    PassengerEnabled off
+  </IfModule>
   Options +FollowSymLinks +Indexes
 <%- if scope.function_versioncmp([@apache_version, '2.4']) >= 0 -%>
   Require all granted
@@ -18,6 +20,8 @@ Alias /pub /var/www/html/pub
 Include /etc/pulp/vhosts80/*.conf
 
 <Location /pulp>
-  PassengerEnabled off
+  <IfModule mod_passenger.c>
+    PassengerEnabled off
+  </IfModule>
   Options +FollowSymLinks +Indexes
 </Location>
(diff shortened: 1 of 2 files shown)


templates/etc/httpd/conf.d/05-foreman.d/katello.conf.erb
```diff
@@ -6,7 +6,9 @@
 Alias /pub /var/www/html/pub
 
 <Location /pub>
-  PassengerEnabled off 
+<% if @use_passenger %>
```
stbenjam: I think this should be `scope['::katello::use_passenger']`, as it's defined in katello class and called from katello::config

beav:
PR updated per comment, and syntax suggestion altered in other conf file as well so they use the same style

stbenjam:
Oh, scoping it is going to break katello_devel: https://github.com/Katello/puppet-katello_devel/blob/master/manifests/apache.pp#L25

I'm surprised this was working at all, as referencing `@variable` shouldn't work in puppet 4 outside of the scope it's in.

stbenjam:
We actually solved that in katello_devel puppet 4 PR (https://github.com/Katello/puppet-katello_devel/pull/80) by moving a copy of the template to the module.  Ideally I think I want to make katello_devel just part of katello module itself, so maybe in the interim we do the same thing with this - copy it to katello_devel :-\

beav:
@stbenjam another option is to do something like

```
<IfModule mod_passenger.c>
  EnablePassenger off
</IfModule>
```

That way it doesn't have to be parameterized at all, or copied between modules.

ekohl:
I agree the puppet-katello_devel situation is ugly and should be fixed. IMHO it shouldn't be an extension of puppet-katello and just call it with the right parameters.

As a first step I'm leaning to writing a foreman::plugin::katello that *just* sets up katello. If you look at it, that's a lot of puppet-katello. The qpid integration is also something I consider part of configuring katello, but setting up Apache to serve pulp is something I'd expect in the puppet-pulp module.

ekohl:
@beav and I certainly like solving it natively in apache. That's the cleanest way of doing it.

beav:
updated to use apache module detection, so code is same whether passenger is enabled or not

ekohl:
ACK on green tests

PR #138, opened 2016-08-08
Fixes #16008 - allow setting puppet_repo_root

beav:
Previously, this setting was not set, which made it default to
`/etc/puppet/environments` in Katello.

This patch adds an explicit setting for `puppet_repo_root`, but
defaults to `/etc/puppet/environments/`. Users can optionally override,
for example to `/etc/puppetlabs/code/environments`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -31,6 +31,8 @@
 #
 # $proxy_password::     Proxy password for authentication
 #
+# $puppet_repo_root::   Directory to write published puppet modules to
+#
 # $cdn_ssl_version::    SSL version used to communicate with the CDN. Optional. Use SSLv23 or TLSv1
 #
 # $num_pulp_workers::   Number of pulp workers to use
@@ -66,6 +68,7 @@ class katello (
   $proxy_port     = $katello::params::proxy_port,
   $proxy_username = $katello::params::proxy_username,
   $proxy_password = $katello::params::proxy_password,
+  $puppet_repo_root = $katello::params::puppet_repo_root,
   $cdn_ssl_version = $katello::params::cdn_ssl_version,
 
   $package_names = $katello::params::package_names,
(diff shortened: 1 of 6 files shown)


beav:
making an update to autodetect this value

manifests/init.pp
```diff
@@ -31,6 +31,8 @@
 #
 # $proxy_password::     Proxy password for authentication
 #
+# $puppet_repo_root::   Directory to write published puppet modules to
```
ekohl: validate it with validate_absolute_path?

ekohl:
Tests fail on puppet 4, otherwise it looks good.

beav:
[test]

beav:
@ekohl if you have a min, can you run the Puppet 4 tests locally for this repo and see if it works? They are failing for me in travis, but also fail when I run them from master locally. It is strange because they ran successfully last week.

stbenjam:
I'd rather not configure this at the Katello level - as this means the entire environment must be 3 or  4 - no mix. My original proposal in the pad is to add an API end-point on the Smart Proxy to say what version of puppet is being used, and have us dynamically determine the path at publish time for each capsule.  Line 142 has an example.  http://pad-katello.rhcloud.com/p/puppet4

But that falls under content and I think @zjhuntin and @cfouant are working on that?

ekohl:
@beav in puppet 4.6.0 I get some error where the whole variable is an empty string, which I don't understand at all.

@stbenjam I think you're right that this should be something at the proxy level, not at the Katello level. However, I don't know Katello well enough to know if that's possible.

ekohl:
@beav https://github.com/ekohl/puppet-katello/commit/73d064875ad78619991fe863957041bf787ca91b does fix it on at least 4.5.x, but broken on 4.6.0 so I assume that's a puppet issue.

beav:
I got sorted out after a quick meeting with @stbenjam. I mistakenly assumed that `puppet_repo_root` was loaded on the capusle, and forgot that repos on capsules are created by the main Katello server.

Closing PR.

PR #139, opened 2016-08-23
refs #10283 - mark parameters advanced

stbenjam:
Anything other than "Parameters" are not shown by default in Kafo help.
This makes the installer help shorter, and reduces the chance users will
change options that really shouldn't be changed unless you really really
really know what you're doing.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,6 +4,21 @@
 #
 # === Parameters:
 #
+# $enable_ostree::      Boolean to enable ostree plugin. This requires existence of an ostree install.
+#                       type:boolean
+#
+# $proxy_url::          URL of the proxy server
+#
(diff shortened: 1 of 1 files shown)


ekohl:
ACK

ehelms:
So the "advanced" parameters still end up as configurable options through kafo they just aren't in the help output? Is there a way to get a list of these through some advanced kafo directive?

stbenjam:
Yup! `--full-help` tells you everything

ehelms:
Cool - APT

ehelms:
@stbenjam Looks like this needs a rebase

stbenjam:
Rebased

stbenjam:
This sat for a little while, would appreciate just a quick sanity check again

PR #140, opened 2016-08-25
Improve ownership and remove unused code

ekohl:
At some point the bundler file was ensured to be owned by the katello user. This traces back to the start of the module (ae7d4aa1039445bbf130d9951ab36d866d551a3b) but there is no real reason for this. We could not manage the file at all, but by ensuring it's owned by root we improve security for existing users.

We also tighten the ownership on the katello.yaml config file.

It removes the remains of client.conf which are unused since 5687ce1ee1fa74b25b4ba95f7ee817f1e58bd307.


ekohl:
Restarted puppet 4 builds due to other PRs being merged. They should pass now.

ehelms:
ACK - thanks @ekohl !

PR #141, opened 2016-08-25
Remove keepalive settings

ekohl:
https://github.com/theforeman/puppet-foreman/commit/0a6966ce22111b1ce592b5e215a948b3734680e7 introduced these parameters in puppet-foreman so these are duplicate.

It should be noted that puppet-foreman has a very different default (100) from what's removed here (10000). It should also be added in a migration in katello-installer if we decide to do this.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -43,10 +43,6 @@
 # $package_names::      Packages that this module ensures are present instead of the default
 #                       type:array
 #
-# $max_keep_alive::     Maximum number of requests to use for the apache MaxKeepAliveRequests parameter
-#                       on the virtualHost for port 443.
-#                       type: integer
-#
 class katello (
 
   $user = $katello::params::user,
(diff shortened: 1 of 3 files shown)


ehelms:
I don't mind centralizing this - as you point out we will need a migration to update things to 10000. I forget exactly the reason but I feel it has something to do with maximizing our client's performance when hitting the server.

ekohl:
The default in Apache appears to be 100 so I'm leaning to not changing puppet-foreman. If Katello needs it, then katello-installer answers may be a better place. Still, 10k sounds like a very high number of requests to do in a single TCP session.

ehelms:
@ekohl Would you mind opening an installer PR with the update so we could merge the changes simultenously?

ekohl:
@ehelms done, see the related PR

ekohl:
Rebased to resolve the merge conflict

stbenjam:
Thanks @ekohl!

PR #142, opened 2016-08-30
Modulesync update

ehelms:


stbenjam:
`$use_passenger` can be removed from init.pp, not used and causing errors on puppet 4 strict_variables as it's not defined in params.pp.

ekohl:
> $use_passenger can be removed from init.pp, not used and causing errors on puppet 4 strict_variables as it's not defined in params.pp.

Added that to https://github.com/Katello/puppet-katello/pull/140

ehelms:
PR part passing

ekohl:
ACK

PR #147, opened 2016-10-24
Modulesync

stbenjam:
Syncing from https://github.com/Katello/foreman-installer-modulesync/pull/11


stbenjam:
Merged based on approval of foreman-installer-modulesync pr

PR #152, opened 2016-11-10
Add right certs to pulp so it can configure crane correctly

stbenjam:
This should fix bats


stbenjam:
Nevermind, https://github.com/Katello/puppet-capsule/pull/105 fixes it instead.

PR #154, opened 2016-11-17
Fixes #17380: Configure ca_cert_file for Candlepin communication

ehelms:


manifests/init.pp
```diff
@@ -71,6 +71,8 @@
   validate_bool($enable_ostree)
   validate_absolute_path($repo_export_dir)
 
+  $candlepin_ca_cert = $::certs::ca_cert
```
ehelms: I introduced this to keep from referencing the certs module directly in the template.

PR #156, opened 2016-11-20
Remove dependency cycle caused by addition of foreman-service anchor

ehelms:


ehelms:
Still fully testing this but this will require all the changes associated to: https://github.com/Katello/katello/pull/6465

ekohl:
Will this continue working?

PR #159, opened 2016-11-22
Optimize tests

ekohl:
By using on_supported_os we can benefit from ONLY_OS thus reducing our test runtime. The use of verify_exact_contents makes it slightly easier to read.


ekohl:
It looks like the difference is negligible since we only support RH7, C7 and SL7 but ONLY_OS is set to RH7 and SL7.

ehelms:
@ekohl did you still want this merged?

ekohl:
@ehelms yes

ekohl:
I rebased this and when the tests are green I intend to merge this.

PR #162, opened 2017-01-09
Refs #16253 - Add max speed var to Katello

chris1984:


manifests/init.pp
```diff
@@ -15,6 +15,8 @@
 #
 # $proxy_password::     Proxy password for authentication
 #
+# $max_speed::          maximum bandwidth used per download in bytes/sec
```
ekohl: To me this was a bit ambiguous. Maybe ```pulp_max_speed``` would be clearer? Should this also be an ```Optional[Integer]```?

ekohl: Also just realised: is this the download that pulp starts or the download speed clients can get?

chris1984:
@stbenjam I used puppet-lint to debug why the tests were failing after running them on my laptop and I am getting this:

WARNING: top-scope variable being used without an explicit namespace on line 128

pulp_max_speed         => $max_speed,

I tried messing around with naming etc, but I am stuck and could use a pointer. Even following this didnt help:

http://puppet-lint.com/checks/variable_scope/

manifests/init.pp
```diff
@@ -122,6 +125,7 @@
     proxy_port             => $proxy_port,
     proxy_username         => $proxy_username,
     proxy_password         => $proxy_password,
+    pulp_max_speed         => $max_speed,
```
ekohl: This should be ```$pulp_max_speed```.

manifests/init.pp
```diff
@@ -15,6 +15,8 @@
 #
 # $proxy_password::     Proxy password for authentication
 #
+# $pulp_max_speed::     Maximum bandwidth used per download in bytes/sec from CDN
```
ekohl: Is this an integer? Can it be ```type:Optional[Integer]```?

stbenjam: Similar to my comment on the Pulp PR, if this only applies to yum I think it should indicate that, e.g. pulp_yum_max_speed

manifests/init.pp
stbenjam: Note this is the cause of the error you were asking about

manifests/init.pp
```diff
@@ -19,6 +19,9 @@
 # $proxy_password::     Proxy password for authentication
 #                       type:String
 #
+# $pulp_max_speed::     Maximum bandwidth used per download in bytes/sec from CDN
```
sean797: Small thing but I don't really like `CDN` here, as this will effect the speed of any RPM downloads.

manifests/init.pp
```diff
@@ -136,6 +140,7 @@
     proxy_port             => $proxy_port,
     proxy_username         => $proxy_username,
     proxy_password         => $proxy_password,
+    $pulp_max_speed        => $yum_max_speed,
```
sean797: Other way round - Pulp class accepts $yum_max_speed.

```
yum_max_speed             => $pulp_max_speed,

File: manifests/init.pp [hunk 6]
```diff
@@ -19,6 +19,9 @@
 # $proxy_password::     Proxy password for authentication
 #                       type:String
 #
+# $pulp_max_speed::     Maximum bandwidth used per download in bytes/sec from CDN
+#                       Type:Integer[Optional]
```
ekohl: I believe this should be ```type:Optional[Integer]```

sean797: Not if its going to use https://github.com/Katello/puppet-pulp/pull/189#issuecomment-273553092 ;)

manifests/init.pp
```diff
@@ -136,6 +140,7 @@
     proxy_port             => $proxy_port,
     proxy_username         => $proxy_username,
     proxy_password         => $proxy_password,
+    yum_max_speed          => $pulp_max_speed,
```
chris1984: @sean797 it looks like tests are still failing with this switched around. I tried using a $ in fron of yum_max_speed and still tests fail.

sean797: Yup, that looks good to me! 👍

manifests/init.pp
```diff
@@ -19,6 +19,9 @@
 # $proxy_password::     Proxy password for authentication
 #                       type:String
 #
+# $pulp_max_speed::     Maximum bandwidth used per download in kb/sec per worker
+#                       Type:Integer[Optional]
```
sean797: Like i said in the other PR - this needs to be a string not an Integer now.

chris1984:
@sean797 @ekohl @stbenjam Thanks for the help everyone, can I get a rereview please now that the other PR is merged tests are passing.

manifests/init.pp
```diff
@@ -19,6 +19,9 @@
 # $proxy_password::     Proxy password for authentication
 #                       type:String
 #
+# $pulp_max_speed::     Maximum bandwidth used per download in kb/sec per worker
```
sean797: Can this match https://github.com/Katello/puppet-pulp/pull/189/files#diff-60ae41fd0a31977447947f59940ee9a4R206 ?

ekohl: Actually this isn't quite correct. ```to_bytes``` without a suffix is still in bytes, but it also converts kB, MB, GB. It looks case sensitive. It's also the SI KB so 1000 bytes rather than the KiB which is 1024 bytes. Not sure how much detail you need to go into, but I can see users being surprised by this.

sean797: aha @ekohl got their first 👍

chris1984:
@ekohl @sean797 does this wording look better?

sean797:
I think we need to get per second in their somewhere?

chris1984:
@sean797 @ekohl Can I get another review please :) Hopefully this is the last one

chris1984:
@stbenjam May I get a please rereview from you please and if all looks well a merge? Thanks

manifests/init.pp
```diff
@@ -19,6 +19,9 @@
 # $proxy_password::     Proxy password for authentication
 #                       type:Optional[String]
 #
+# $pulp_max_speed::     The maximum download speed per second for a Pulp task, such as a sync. (e.g. "4 Kb" (Uses SI KB), 4MB, or 1GB" )
+#                       Type:String
```
stbenjam: This should be a lowecase `t` to match the others

chris1984:
@stbenjam Thanks for the catch, moved to lowercase t. Can I get a rereview and merge if it looks good please. Thanks.

ehelms:
This is currently the cause of pipeline failures -- http://ci.theforeman.org/view/Katello%20Pipeline/job/release_test_katello_nightly_el7/170/console

stbenjam:
@ehelms https://github.com/Katello/puppet-katello/pull/165

The param types really should have some testing against their default values.

PR #170, opened 2017-03-15
Preparations to seperate logic out of the certs module

Klaas-:
the qpid exchange creation should not be part of the puppet-certs module

additional change to make the host it uses for the qpid connection as parameter, defaults to localhost

someone with some more qpid insight should determine if that exchange is actually needed and used please :)

### DIFF
--- a/manifests/qpid.pp
+++ b/manifests/qpid.pp
@@ -2,14 +2,22 @@
 class katello::qpid (
   $client_cert,
   $client_key,
-  $katello_user          = $::katello::user,
-  $candlepin_event_queue = $::katello::candlepin_event_queue,
+  $katello_user             = $::katello::user,
+  $candlepin_event_queue    = $::katello::candlepin_event_queue,
+  $candlepin_qpid_exchange  = $::katello::candlepin_qpid_exchange,
 ){
   if $katello_user == undef {
     fail('katello_user not defined')
   } else {
     Group['qpidd'] ->
     User<|title == $katello_user|>{groups +> 'qpidd'}
+  }
+  exec { 'create candlepin qpid exchange migrate':
+    command   => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://localhost:5671' add exchange topic ${candlepin_qpid_exchange} --durable",
+    unless    => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://localhost:5671' exchanges ${candlepin_qpid_exchange}",
+    path      => '/usr/bin',
+    require   => Service['qpidd'],
+    logoutput => true,
   }
   exec { 'create katello entitlements queue':
     command   => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://localhost:5671' add queue ${candlepin_event_queue} --durable",
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -63,5 +63,6 @@ class katello::params {
 
   $qpid_url = 'amqp:ssl:localhost:5671'
   $candlepin_event_queue = 'katello_event_queue'
+  $candlepin_qpid_exchange = 'event'
   $enable_ostree = false
 }


ehelms:
I think this makes sense once tests are passing.

Klaas-:
@ehelms yeah I forgot to include my puppet-git-hooks ;) anyways once you decide this is mergeable I'll create another pr to remove the resource in puppet-certs, they should be merged at the same time :)

Klaas-:
the name is temporary because otherwise the tests fail, I'll rename it to the old one once its deemed mergable

manifests/qpid.pp
```diff
@@ -2,25 +2,34 @@
...
     Group['qpidd'] ->
     User<|title == $katello_user|>{groups +> 'qpidd'}
   }
+  exec { 'create candlepin qpid exchange migrate':
+    command   => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://${qpid_hostname}:5671' add exchange topic ${candlepin_qpid_exchange} --durable",
+    unless    => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://${qpid_hostname}:5671' exchanges ${candlepin_qpid_exchange}",
+    path      => '/usr/bin',
+    require   => Service['qpidd'],
+    logoutput => true,
+  } ~>
```
ekohl: Is this notification needed? There doesn't appear to be a relation from the arguments used.

Klaas-: @ekohl I doubt its needed, but I wanted to get this out of puppet-certs before changing to the new qpid exchange type and fixing other stuff :)

ekohl:
Needs a rebase now.

Klaas-:
@ekohl  rebased and I took out the hostname parameter I wanted to introduce, will do that #171

manifests/qpid.pp
```diff
@@ -2,15 +2,23 @@
...
     Group['qpidd'] ->
     User<|title == $katello_user|>{groups +> 'qpidd'}
   }
+  exec { 'create candlepin qpid exchange migrate':
+    command   => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://localhost:5671' add exchange topic ${candlepin_qpid_exchange} --durable",
+    unless    => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://localhost:5671' exchanges ${candlepin_qpid_exchange}",
+    path      => '/usr/bin',
+    require   => Service['qpidd'],
+    logoutput => true,
+  } ~>
```
ekohl: I'm not sure this has to be a notify

Klaas-: ACK. from logic I'd say it does not need any anchors

Klaas-:
I think these are for communication between katello and candlepin but maybe @ehelms can elaborate. I'll change the commit message :)

ehelms:
Ok, after some digging around, here is how I believe things should be done.

Candlepin relies on an 'event' exchange to exist to push events out on to. This is the only requirement from Candlepin as far as I can tell and can be seen via [1].

Katello wishes to create a queue and then bind events to that queue for consumption. The events are coming from the exchange that Candlepin creates and is "managing". While Katello manages the queue and the events subscribed to on the queue. Thus the two are separated in logic except that Katello needs to know the name of the exchange Candlepin has created and is using so as to bind the queue and events to it.

In the good work @chris1984 did to create a generic type on `puppet-qpid` for binding events, there as a subtely that was missed that I feel is important when trying to decipher this. That being that `event` is a specific exchange created by Candlepin (in reality our puppet module) for publishing of events. And that to perform a qpid `bind` action, one needs both the queue and the exchange to bind to.

Thus, I have opened [2] to make this Qpid action generic per the documentation [3] and not tie it to a specific exchange as I think that could prevent confusion trying to reuse it. As @ekohl pointed out, I do think `puppet-candlepin` should probably be creating the exchange since it is the publisher. And `puppet-katello` should be creating the queue and event bindings since it is the consumer.

[1] https://github.com/candlepin/candlepin/blob/2150cfc384c210ba09f7aa40acb8de3c0ef42d79/server/bin/qpid/configure-qpid.sh#L222
[2] https://github.com/Katello/puppet-qpid/pull/52
[3] https://qpid.apache.org/releases/qpid-0.32/cpp-broker/book/chapter-Managing-CPP-Broker.html

ekohl:
@ehelms thanks for digging that up. It confirms what I suspected and I'm glad we're on the same page where responsibilities of modules lie.

Klaas-:
@ehelms 
so close this and make a a PR against puppet-candlepin once Katello/puppet-qpid#52 is introduced?

ekohl:
@Klaas- I think that would be the best solution yes.

PR #172, opened 2017-03-18
refresh services when certs change

timogoebel:
This commits adds code to refresh services when certs change.
Previously, this has been done in puppet-certs module.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -169,6 +169,8 @@ class katello (
   class { '::certs::foreman': }
 
   Exec['cpinit'] -> Exec['foreman-rake-db:seed']
+  Class['certs::candlepin'] ~> Service['tomcat']
+  Class['certs::qpid'] ~> Service['qpidd']
 
   User<|title == apache|>{groups +> $user_groups}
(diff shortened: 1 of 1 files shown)


timogoebel:
... we might want to wait for https://github.com/Katello/puppet-katello/pull/171/files before merging this.

manifests/init.pp
```diff
@@ -171,4 +171,11 @@
   Exec['cpinit'] -> Exec['foreman-rake-db:seed']
 
   User<|title == apache|>{groups +> $user_groups}
+
+  # lint:ignore:spaceship_operator_without_tag
+  Certs::Ssltools::Certutil <| |> ~> Service['qpidd']
```
ekohl: Can't we catch this in some other way? I'd think the ```Class['certs::qpid']``` would also take care of this.

timogoebel:
Updated in the hope, that it'll fix the pipeline.

ekohl:
@ehelms I am

ehelms:
Thanks @timogoebel !

PR #175, opened 2017-03-22
Fixes #18812 - Add exec to delete unfiltered queue

chris1984:


timogoebel:
... what happens if I want the queue to listen on `*.*`?

manifests/qpid.pp
```diff
@@ -11,6 +11,13 @@
     Group['qpidd'] ->
     User<|title == $katello_user|>{groups +> 'qpidd'}
   }
+  exec { 'delete katello entitlements queue if bound to *.*':
+    command   => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://localhost:5671' del queue ${candlepin_event_queue} --force",
+    onlyif    => "qpid-config --ssl-certificate ${client_cert} --ssl-key ${client_key} -b 'amqps://localhost:5671' list binding | grep ${candlepin_event_queue} | grep '*.*'",
+    path      => '/usr/bin',
+    require   => Service['qpidd'],
+    logoutput => true,
+  }
```
stbenjam: You need a `~>` so the creation is done after this

stbenjam:
@timogoebel You don't.

We moved to limiting the queues to only be bound to certain events, this PR ensures we delete the queue if it has the old setup, and recreate it.

https://github.com/Katello/puppet-katello/pull/175/files#diff-d156c37507c1a892d1de9c1db261915eR28

stbenjam:
I think I saw on one of the other PR's, you want to move this to puppet-candlepin which would be OK. In that case, supporting non-Katello candlepin users, they theoretically might be interested in other events or all of them, which means parameterization of the bindings, and more complicated logic here.  Is that why you were asking?

Given there are no non-Katello/non-Red Hat users of Candlepin AFAIK, I'm not sure we need to worry about that in the short-term.

timogoebel:
@stbenjam:

> @timogoebel You don't.

Fair enough.

> you want to move this qpid class to puppet-candlepin

Actually, I don't. What I want is to be able is to have quidd run on a different server then candlepin. That's why I asked. But if the queue isn't required at all, go a head. Sorry for the noise. Thanks for the clarification. :-)

manifests/qpid.pp
ekohl: Or just a ```->```.

ekohl:
>> you want to move this qpid class to puppet-candlepin
> 
> Actually, I don't. What I want is to be able is to have quidd run on a different server then candlepin. That's why I asked. But if the queue isn't required at all, go a head. Sorry for the noise. Thanks for the clarification. :-)

That was my question. I don't know if it should be moved to puppet-candlepin but I was wondering whose responsibility it is to set up these exchanges and event binding: Candlepin or Katello. If candlepin can't function without something, it belongs in puppet-candlepin. If it's for Katello integration, then it's most likely puppet-katello (in ```katello::candlepin```). Possibly it could be an optional ```candlepin::katello``` class, but not sure that's desired.

ekohl:
Also 👍 for doing this in the puppet module rather than the installer.

manifests/qpid.pp
stbenjam: Don't we need the notify? I thought if we deleted the queue, we should cause the other one to refresh.  Maybe if just the ordering is right it works because of the onlyif/unless on the execs.

stbenjam:
Oh ok sorry for the distraction, I thought I saw a discussion about that on one of the recent PR's. 

Anyway, having candlepin talk to a queue is a feature, but I think candlepin still works without it.  I'm not sure which class it belongs in but we can deal with that separately.

manifests/qpid.pp
stbenjam: Er, nevermind, you're right, the ordering is enough, so just `->`

chris1984:
@stbenjam @ekohl added the "->" if you want to review again when you get a moment. Did not know if you wanted to keep it here in this repo or move it elsewhere but happy to do whatever you guys want. Tests are green.

ekohl:
@stbenjam given https://github.com/Katello/puppet-katello/pull/170#issuecomment-288478211 I think everything related to the event queue belongs in puppet-katello while the exchange creation belongs in puppet-candlepin.

stbenjam:
@ekohl That sounds fine, but we can deal with that separately from this issue.  The changes here LGTM.

stbenjam:
Thanks @chris1984

PR #179, opened 2017-03-28
Specify qpid SSL key to qpid commands

ehelms:
@ekohl needed Katello changes for your key specification


manifests/qpid.pp
```diff
@@ -17,15 +17,18 @@
     command  => "del queue ${candlepin_event_queue} --force",
     onlyif   => "list binding | grep ${candlepin_event_queue} | grep '*.*'",
     ssl_cert => $client_cert,
+    ssl_key  => $client_key,
   } ->
   qpid::config_cmd { 'create katello entitlements queue':
```
ekohl: This can also be a ```qpid::config::queue``` now.

ekohl: Ignore that, we don't have a queue define yet. That's exchange.

PR #182, opened 2017-03-30
fixes #19097 - remove /subscription route, change file headers

stbenjam:


ekohl:
Lint fails for unrelated reasons. I like the simplification.

PR #185, opened 2017-04-04
Add optional repository management

ekohl:


manifests/init.pp
```diff
@@ -64,6 +64,20 @@
 # $package_names::      Packages that this module ensures are present instead of the default
 #                       type:Array[String]
 #
+# $manage_repo::        Whether to manage the yum repository
+#                       type:Boolean
+#
+# $repo::               Which yum repository to install. For example
```
timogoebel: I think, this name is misleading. Why not simply `version`?

manifests/repo.pp
```diff
@@ -0,0 +1,17 @@
+class katello::repo (
+  $manage_repo = $::katello::manage_repo,
+  $repo        = $::katello::repo,
+  $dist        = $::katello::repo_yumcode,
+  $gpgcheck    = $::katello::repo_gpgcheck,
+  $gpgkey      = $::katello::repo_gpgkey,
+) {
+  if $manage_repo {
```
timogoebel: Why not move that if to init.pp?

if $manage_repo {
 include ::katello::repo
}

manifests/init.pp
ekohl: I originally copied this from puppet-foreman into puppet-candlepin. We could change it to ```repo_version``` and similar in puppet-candlepin.

manifests/repo.pp
ekohl: Oh I think I forgot the include there. The reason I put it here is that this makes chaining easy

ehelms:
I think this generally looks good after addressing the comments. I believe this is the last PR @ekohl wanted before we released all Katello modules. I'd like to get that done tomorrow if we could so we can build the katello-installer for RC1 and test it. I had mentioned this to @stbenjam since he kindly does a lot of the module releases around this time in the process.

ekohl:
Updated

PR #187, opened 2017-04-07
Make GPG key optional

ehelms:


ehelms:
Fixes http://ci.theforeman.org/job/release_test_katello_nightly_el7/305/artifact/debug/fb-install-katello.bats.tap/*view*/

manifests/init.pp
```diff
@@ -75,7 +75,7 @@
 #                       type:Boolean
 #
 # $repo_gpgkey::        The GPG key to use
-#                       type:String
+#                       type:String[Optional]
```
stbenjam: Other way, `Optional[String]`

ehelms: Doh, need a kafo linter.

ehelms:
Updated

PR #188, opened 2017-04-20
Refactor to Puppet 4 types

stbenjam:

PR #189, opened 2017-04-25
Move to puppet 4

ekohl:
Includes a modulesync and https://github.com/Katello/puppet-katello/pull/188 with some type fixes.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,6 +1,8 @@
 fixtures:
   repositories:
-    stdlib:        "git://github.com/puppetlabs/puppetlabs-stdlib.git"
+    stdlib:
+      repo: "git://github.com/puppetlabs/puppetlabs-stdlib.git"
+      branch: "release"
     extlib:        "git://github.com/puppet-community/puppet-extlib"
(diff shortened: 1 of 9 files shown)


ekohl:
There is no newer puppet-foreman version out so I think it's weird to increase it.

stbenjam:
I guess, but does this work with the latest released version on the Forge? Tests would be pulling out of git which is 8.0, I feel like the requirements should match what we're testing against

ekohl:
You're right that the version in metadata.json is already 8.0.0. Given we hardly call anything from the foreman module (just ```foreman::config::passenger::fragment``` and some resource ordering) I think it's safe to bump the requirement because those are unlikely to change.

ekohl:
I've added 3 additional commits, 2 are cleanups I think are safe.

ekohl:
I forgot that there are more entry classes, like admin and crane. Fixing those as well.

ekohl:
@stbenjam can you re-review?

ekohl:
Merged without the stdlib commit because I trust it'll become stable again and shouldn't impact releases.

PR #190, opened 2017-05-18
Simplify variables

ekohl:


chris1984:
@ekohl what is the "::" doing is reading the entire scope?

ekohl:
@chris1984 absolute variables/global scope. In the past Puppet had weird scoping rules in that if you used ```$katello::var``` inside a class ```profiles``` then it would also look for ```$::profiles::katello::var```. This could lead to subtle bugs. I think Puppet 4 changed the scoping so it may no longer be necessary but it's still something I try to do. Partly because I like that it's explicitly a variable from a different class/scope.

manifests/init.pp
```diff
@@ -24,6 +24,8 @@
 #
 # $group::              The Katello system user group
 #
+# $deployment_url::     The URL katello is deployed under
```
ehelms: This is not exactly what this is intended for and I don't think we should expose this. This is something we have to configure for Candlepin and is specific to our use case.

ekohl: I exposed this because puppet-lint complained otherwise but you're right that users might override this which could be bad. I'll fix this up.

ekohl:
Updated

ekohl:
With https://github.com/Katello/puppet-katello/pull/204 this is no longer relevant.

PR #191, opened 2017-05-18
Split apache fragments

ekohl:
This patch drops support for Apache 2.2, but we only support EL7 which
ships 2.4.

It is untested, but I need this to easily test it in a pipeline run.


ekohl:
I believe this is ready to be merged now.

PR #192, opened 2017-05-25
Fixes #19667 - expose Candlepin DB setup

mbacovsky:
Exposing Candlpin DB parameters via katello module allows users to
customize the DB setup using installer parameters.

Current default Candlepin DB setup remains unaffected. Using external
Postgress with prepared DB is possible.

I've tested the default Katello install and also fresh installation with remote DB using:
```
foreman-installer --scenario katello --foreman-admin-password changeme --foreman-db-host X.X.X.X --foreman-db-password foreman --foreman-db-database foreman --katello-candlepin-db-host X.X.X.X - -katello-candlepin-db-name candlepin --katello-candlepin-db-password candlepin
``` 
The remote DB use-case assumes 
 - the DB server allows connection from the Katello server 
 - has one database for Foreman and one for Candlepin
 - postgress users owning the databases can auth in using md5

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -69,6 +69,8 @@
 #
 # $candlepin_db_ssl_verify:: Boolean indicating if the SSL connection to the database should be verified
 #
+# $candlepin_manage_db:: Boolean indicating whether a database should be installed, this includes db creation and user
+#
 class katello (
   String $user = $::katello::params::user,
   String $group = $::katello::params::group,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -54,6 +54,16 @@
 #
 # $repo_gpgkey::        The GPG key to use
 #
+# $candlepin_db_host::  Host with Candlepin DB
+#
+# $candlepin_db_port::  Port accepting connections to Candlepin DB
+#
+# $candlepin_db_name::  Name of the Candlepin DB
+#
+# $candlepin_db_user::  Candlepin DB user
+#
+# $candlepin_db_password:: Candlepin DB password
+#
 class katello (
   String $user = $::katello::params::user,
   String $group = $::katello::params::group,
@@ -83,6 +93,12 @@ class katello (
   String $repo_version = $::katello::params::repo_version,
   Boolean $repo_gpgcheck = $::katello::params::repo_gpgcheck,
   Optional[String] $repo_gpgkey = $::katello::params::repo_gpgkey,
+
+  String $candlepin_db_host = $::katello::params::candlepin_db_host,
+  Optional[String] $candlepin_db_port = $::katello::params::candlepin_db_port,
+  String $candlepin_db_name = $::katello::params::candlepin_db_name,
+  String $candlepin_db_user = $::katello::params::candlepin_db_user,
+  String $candlepin_db_password = $::katello::params::candlepin_db_password,
 ) inherits katello::params {
   $candlepin_ca_cert = $::certs::ca_cert
   $pulp_ca_cert = $::certs::katello_server_ca_cert
@@ -120,6 +136,11 @@ class katello (
     amqp_truststore              => $::certs::candlepin::amqp_truststore,
     qpid_ssl_cert                => $::certs::qpid::client_cert,
     qpid_ssl_key                 => $::certs::qpid::client_key,
+    db_host                      => $katello::candlepin_db_host,
+    db_port                      => $katello::candlepin_db_port,
+    db_name                      => $katello::candlepin_db_name,
+    db_user                      => $katello::candlepin_db_user,
+    db_password                  => $katello::candlepin_db_password,
   } ~>
   class { '::certs::qpid_client': } ~>
   class { '::pulp':
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -106,6 +106,7 @@ class katello (
   String $candlepin_db_password = $::katello::params::candlepin_db_password,
   Boolean $candlepin_db_ssl = $::katello::params::candlepin_db_ssl,
   Boolean $candlepin_db_ssl_verify = $::katello::params::candlepin_db_ssl_verify,
+  Boolean $candlepin_manage_db = $::katello::params::candlepin_manage_db,
 ) inherits katello::params {
   $candlepin_ca_cert = $::certs::ca_cert
   $pulp_ca_cert = $::certs::katello_server_ca_cert
@@ -150,6 +151,7 @@ class katello (
     db_password                  => $candlepin_db_password,
     db_ssl                       => $candlepin_db_ssl,
     db_ssl_verify                => $candlepin_db_ssl_verify,
+    manage_db                    => $candlepin_manage_db,
   } ~>
   class { '::certs::qpid_client': } ~>
   class { '::pulp':
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -95,7 +95,7 @@ class katello (
   Optional[String] $repo_gpgkey = $::katello::params::repo_gpgkey,
 
   String $candlepin_db_host = $::katello::params::candlepin_db_host,
-  Optional[String] $candlepin_db_port = $::katello::params::candlepin_db_port,
+  Optional[Integer[0, 65535]] $candlepin_db_port = $::katello::params::candlepin_db_port,
   String $candlepin_db_name = $::katello::params::candlepin_db_name,
   String $candlepin_db_user = $::katello::params::candlepin_db_user,
   String $candlepin_db_password = $::katello::params::candlepin_db_password,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -64,6 +64,11 @@
 #
 # $candlepin_db_password:: Candlepin DB password
 #
+# $candlepin_db_ssl::   Boolean indicating if the connection to the database should be over
+#                       an SSL connection.
+#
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -120,6 +136,11 @@
     amqp_truststore              => $::certs::candlepin::amqp_truststore,
     qpid_ssl_cert                => $::certs::qpid::client_cert,
     qpid_ssl_key                 => $::certs::qpid::client_key,
+    db_host                      => $katello::candlepin_db_host,
```
sean797: What do you think about getting these from `$::candlepin::db_*` params ? I think it should mean we wouldn't have to set defaults here and in puppet-candlepin.

ekohl: Defaulting to undef would remove the need to copy defaults. That does mean changing the types to Optional. Additionally all these variables are already in the local scope so you can use ```$candlepin_db_host```.

manifests/init.pp
```diff
@@ -83,6 +93,12 @@
   String $repo_version = $::katello::params::repo_version,
   Boolean $repo_gpgcheck = $::katello::params::repo_gpgcheck,
   Optional[String] $repo_gpgkey = $::katello::params::repo_gpgkey,
+
+  String $candlepin_db_host = $::katello::params::candlepin_db_host,
+  Optional[String] $candlepin_db_port = $::katello::params::candlepin_db_port,
```
ekohl: In the candlepin module this is an ```Optional[integer[0, 65535]]```.

manifests/init.pp
ekohl: I now see you want to use these values in the installer so copying makes sense.

mbacovsky: @sean797 it seems neither the katello class nor the katello::params can see the the db_name's default value from  candlepin. Is there any trick how to do that?

mbacovsky: It is worth mentioning that we need to have the defaults in the class declaration for the installer to see it and display in the help.

manifests/init.pp
```diff
@@ -95,7 +95,7 @@
   Optional[String] $repo_gpgkey = $::katello::params::repo_gpgkey,
 
   String $candlepin_db_host = $::katello::params::candlepin_db_host,
-  Optional[String] $candlepin_db_port = $::katello::params::candlepin_db_port,
+  Optional[integer[0, 65535]] $candlepin_db_port = $::katello::params::candlepin_db_port,
```
ekohl: This should be Integer (capital I)

manifests/init.pp
ekohl: @mbacovsky I don't think that's worth it. They're static in candlepin, except the password but that's handled by using the same password cache

ekohl: Could you change the variables to ```$candlepin_*```? They're already in local scope.

mbacovsky:
@ekohl, updated, variables scoping fixed.  There are also other parameters with the `$katello::` module prefix.
```ruby
   class { '::candlepin':
    user_groups                  => $katello::user_groups,
    oauth_key                    => $katello::oauth_key,
```
Is there a reason for that or should I fix it to meke it consistent?

ekohl:
> Is there a reason for that or should I fix it to meke it consistent?

https://github.com/Katello/puppet-katello/pull/190 should fix those.

mbacovsky:
@ekohl  cool, thanks!

ekohl:
@mbacovsky we generally have the rule that you can merge your own PRs if there's an ack from another maintainer. So if you want you can merge this now.

ehelms:
A couple things to consider, this locks in to `postgresql` as `db_type` is considered in the conf (https://github.com/Katello/puppet-candlepin/blob/master/templates/_candlepin_database.conf.erb#L3). Also, I think this would only support non-ssl external databases today which seems like a bad thing once you go external (https://github.com/Katello/puppet-candlepin/blob/master/manifests/params.pp#L12).

ekohl:
@ehelms previously the parameters were not exposed either and could only be set through hiera parameters. This doesn't change but it's a good point to address. If that is exposed, then the katello-installer hook should be modified to take advantage as well.

mbacovsky:
@ehelms, we do not support other DB backends currently and there would be much more needed to make it usable and I'd prefer not to touch it now.
For the SSL support I'll need to do some checks and update the PR (and other places accordingly). Thanks for pointing it out.

mbacovsky:
@ehelms I've tested Katello/Satellite with SSL required by postgres config. Rails can do it by default with fallback to not verify CA. Candlepin need ssl=true in jdbc config and the postgres CA in system trust or the verification explicitly off. I've updated this PR to include the SSL options.  

To have SSL supported katello-installer needs to be patched, PR will follow.

ehelms:
I think we would need to also include the `manage_db` (https://github.com/Katello/puppet-candlepin/blob/master/manifests/database/postgresql.pp#L27) parameter for this to work as intended. Otherwise, a database and postgres server are created/installed and started on the box.

mbacovsky:
@ehelms, `manage_db` added.

manifests/init.pp
```diff
@@ -83,6 +98,15 @@
...
   Optional[String] $repo_gpgkey = $::katello::params::repo_gpgkey,
+
+  String $candlepin_db_host = $::katello::params::candlepin_db_host,
+  Optional[Integer[0, 65535]] $candlepin_db_port = $::katello::params::candlepin_db_port,
+  String $candlepin_db_name = $::katello::params::candlepin_db_name,
+  String $candlepin_db_user = $::katello::params::candlepin_db_user,
+  String $candlepin_db_password = $::katello::params::candlepin_db_password,
+  Boolean $candlepin_db_ssl = $::katello::params::candlepin_db_ssl,
+  Boolean $candlepin_db_ssl_verify = $::katello::params::candlepin_db_ssl_verify,
+  Boolean $candlepin_manage_db = $::katello::params::candlepin_manage_db,
```
ehelms: Missing docs

ehelms:
I ran through a test of this and got a successful installation with no postgres database started on the server itself. However, I did hit a bats tests failure that I am unclear if is related to this separation or not. The error I get:

```
not ok 3 create a product
# (in test file fb-content-katello.bats, line 29)
#   `hammer product create --organization="${ORGANIZATION}" --name="${PRODUCT}" | grep -q "Product created"' failed
# Could not create the product:
#   Task d65fdf10-b1ab-4196-837b-b97b3751324f: ActiveRecord::RecordNotFound: Couldn't find User with 'id'=4 [WHERE (1=0)]
ok 4 create package repository
```

The testing scenario I am using is the located at [1]. I have crafted that scenario based on this PR, and whats available for Foreman so please double check that I am setting up a proper configuration with the right values. The installation did succeed, but bats tests failures make me pause.


[1] https://github.com/theforeman/forklift/pull/269

ehelms:
Another item I noticed, that needs to be considered is with postgres no longer on the server, things like `katello-service status` will indicate failure since postgres is not on the box. Does katello-backup and restore continue to work with an external database?

beav:
I think `katello-service` will work ok, as long as `systemctl is-enabled postgres` returns nonzero. We may need to mask the service or something, not sure.

mbacovsky:
@ehelms, thanks for testing this. I've seen the bats error on the list today (http://projects.theforeman.org/issues/20040) -> http://projects.theforeman.org/issues/20040 so probably not related.

And yes, katello-backup and katello-service needs to be reviewed/updated and are on my TODO list next.

manifests/init.pp
```diff
@@ -54,6 +54,23 @@
...
+#
+# $candlepin_db_port::  Port accepting connections to Candlepin DB
+#
+# $candlepin_db_name::  Name of the Candlepin DB
+#
+# $candlepin_db_user::  Candlepin DB user
+#
+# $candlepin_db_password:: Candlepin DB password
+#
+# $candlepin_db_ssl::   Boolean indicating if the connection to the database should be over
```
sean797: What about adding "Requires DB host's CA Cert in the system trust" or something similar ?

mbacovsky:
@sean797, thanks, doc added

NeilHanlon:
Fwiw, I used this (and related) PRs to upgrade to 3.4 with limited issues. Only big thing I'm seeing is that it does not seem to actually respect the SSL boolean, as my candlepin.conf was setup with `?ssl=true` and setting the param in the CLI just caused errors (regarding parsing a string to a bool, etc). Unfortunately, I neglected to record all these.

Additionally, I had issues applying this patch specifically against the distributed code because of a mismatch with trunk vs distribution (there was a param to the module removed which caused `patch` to have issues). I will look through my logs to find more about this if necessary (it was also discussed in #theforeman-dev)

NeilHanlon:
@beav correct

beav:
Would it be OK to merge this, and then check for issues once it's in nightly? The issues mentioned in https://github.com/Katello/puppet-katello/pull/192#issuecomment-311383732 might be from differences in the patch between master and 3.4.

mbacovsky:
@beav  I'll try to reproduce this tomorrow and let you know. If we can wait with merging this till tomorrow to make sure we didn't miss anything important it would be great.

mbacovsky:
@beav, @NeilHanlon:  I was not able to reproduce the problem with cleanly applied patches. It correctly removed the ssl=true from my setup. So I don't found any obstacles for merging it.

beav:
@mbacovsky merged, thanks!

PR #194, opened 2017-06-06
Remove crane_data_dir

jlsherrill:
as it should be handled by foreman_proxy_content, that pulls
in crane.


jlsherrill:
note sure how i opened this twice :)

PR #198, opened 2017-07-18
Allow pulp 5.x

ekohl:


chris1984:
@ekohl is there any harm with bringing in Pulp 5.x

ekohl:
Currently I don't think there is. Everything should be compatible but it dropped Puppet 3. Since Katello did the same that shouldn't be an issue.

PR #200, opened 2017-08-03
Refs #19514 - Expose qpid params to user

chris1984:


chris1984:
Set the max at 5000 per this Jira:
https://issues.apache.org/jira/browse/QPID-7677

manifests/init.pp
```diff
@@ -63,6 +67,8 @@
   String $oauth_secret = $::katello::params::oauth_secret,
 
   String $post_sync_token = $::katello::params::post_sync_token,
+  Integer[0, 5000] $qpid_session_unacked = $::katello::params::qpid_session_unacked,
+  Integer[0, 1000] $qpid_wcache_page_size = $::katello::params::qpid_session_unacked,
```
ehelms: Right hand side is  copy/paste error

manifests/init.pp
```diff
@@ -18,6 +18,10 @@
 #
 # $repo_export_dir::    Directory to create for repository exports
 #
+# $qpid_session_unacked:: Buffer if the broker has a large number of sessions and the memory overhead is a problem
+#
+# $qpid_wcache_page_size:: The size (in KB) of the pages in the write page cache
```
ehelms: I think these should be advanced parameters.

manifests/params.pp
```diff
@@ -35,6 +35,10 @@
   # Pulp max speed setting
   $pulp_max_speed = undef
 
+  # Qpid perf settings
+  $qpid_session_unacked = 10
```
ekohl: The qpid in EPEL doesn't accept this parameter and failing to start up. I think we should revert this.

PR #201, opened 2017-08-07
fixes #20353 - Runs apipie cache after ostree

parthaa:
We need to run ```/usr/sbin/foreman-rake apipie:cache:index ``` after installing ostree. So added that option in the installer and closing https://github.com/Katello/katello-packaging/pull/483

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -7,7 +7,7 @@ class katello::install {
   if $katello::enable_ostree {
     package { $katello::rubygem_katello_ostree:
       ensure => installed,
-      notify => Service['foreman-tasks', 'httpd'],
+      notify => [Service['foreman-tasks', 'httpd'], Exec['foreman-rake-apipie:cache:index']],
     }
   }
 }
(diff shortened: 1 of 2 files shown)


manifests/config.pp
```diff
@@ -45,4 +45,15 @@
...
+    exec { "foreman-rake-ostree-apipie":
+      command     => "/usr/sbin/foreman-rake apipie:cache:index && touch /usr/share/foreman/tmp/ostree_apipie_index_done",
+      user        => $::foreman::user,
+      environment => sort(join_keys_to_values({'HOME' => $::foreman::app_root}, '=')),
+      logoutput   => 'on_failure',
+      refreshonly => true,
+      creates => '/usr/share/foreman/tmp/ostree_apipie_index_done',
+      require  => Package[$katello::rubygem_katello_ostree],
+    }
+  }
```
ehelms: This seems like overkill. Could you now in https://github.com/Katello/puppet-katello/blob/master/manifests/install.pp#L10 add a notify to `Exec['foreman-rake-apiepie:cache:index']` ?

parthaa:
@ehelms @ekohl I 'll looking to addressing your notify comment. But I had one question. Were you guys ok with my idea of adding a 'touch' file to not re-execute this call on every run of the installer?  I have a feeling ```refreshonly => true,``` is enough. What do you guys think?

ehelms:
LGTM - @ekohl do you forsee any gotcha with this?

PR #203, opened 2017-08-09
Fixes #20518 - set pulp tasks to 2

chris1984:
Downstream this change was made for performance reasons, reference in the following BZ:

https://bugzilla.redhat.com/show_bug.cgi?id=1388631

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -30,7 +30,7 @@ class katello::params {
   $proxy_password = undef
 
   $num_pulp_workers = min($::processorcount, 8)
-  $max_tasks_per_pulp_worker = undef
+  $max_tasks_per_pulp_worker = 2
 
   # Pulp max speed setting
   $pulp_max_speed = undef


chris1984:
Downstream this change was made for performance reasons, reference in the following BZ:

https://bugzilla.redhat.com/show_bug.cgi?id=1388631

ekohl:
Where does the number 2 come from? Setting this 2 also means the text (from the bugzilla) "To disable this completely, users can set '--katello-max-tasks-per-pulp-worker' to 'undef'." needs to be changed to ```to '0'```.

chris1984:
@ekohl it comes from https://bugzilla.redhat.com/show_bug.cgi?id=1388631#c19 I can let @beav comment more on this why the change was made as he was doing the perf stuff

beav:
There is nothing special about the number 2, I think the pulp team settled on it after testing and noticing that it's extremely cheap to shut down and spin up a new worker process, so it's done after every other task. You could probably pick another small number and it would work similarly but 2 is the convention. I've never seen issues with 2, but I would most likely say the same if it was any other reasonably small number 😸

PR #204, opened 2017-08-09
[WIP] Split classes & separation of services

ekohl:
This allows deploying parts of the stack while still defaulting to a monolithic all-in-one deployment.


ekohl:
@Klaas- @timogoebel I think this is the better version. I'm still considering adding a parameter to include the meta package but let's first try to get this merged.

manifests/init.pp
```diff
@@ -115,104 +139,59 @@
...
+  Boolean $enable_candlepin = $::katello::params::enable_candlepin,
+  Boolean $enable_qpid = $::katello::params::enable_qpid,
+  Boolean $enable_qpid_client = $::katello::params::enable_qpid_client,
+  Boolean $enable_pulp = $::katello::params::enable_pulp,
+  Boolean $enable_application = $::katello::params::enable_application,
+
+  String $candlepin_hostname = $::katello::params::candlepin_hostname,
+  String $pulp_hostname = $::katello::params::pulp_hostname,
+  String $qpid_hostname = $::katello::params::qpid_hostname,
+  Stdlib::HTTPUrl $candlepin_url = "https://${candlepin_hostname}:8443/candlepin",
```
ekohl: @Klaas- this may work in pure puppet but with the installer this breaks:

```
 | URI::InvalidURIError: bad URI(is not URI?): https://${candlepin_hostname}:8443/candlepin
```

I'd prefer not to split the hostname and URLs if possible. Shall we try with just the segments exposed and compose the URL variables inside the body?

Klaas-: I think we need the qpid hostname for ::candlepin, thats why I split them up, the other two just using url should be fine. I don't recall pulp_hostname and candlepin_hostname being used anywhere except for building the urls

manifests/application.pp
```diff
@@ -0,0 +1,71 @@
+# Install and configure the katello application itself
+class katello::application(
```
timogoebel: Nitpick: Add an extra whitespace between `n` and `(`

manifests/init.pp
Klaas-: maybe we can go the other way around for candlepin qpid hostname and do something like split($qpid_url, ':')[2]

manifests/candlepin.pp
```diff
@@ -0,0 +1,54 @@
...
+  Optional[Integer[0, 65535]] $db_port = $::katello::candlepin_db_port,
+  String $db_name = $::katello::candlepin_db_name,
+  String $db_user = $::katello::candlepin_db_user,
+  String $db_password = $::katello::candlepin_db_password,
+  Boolean $db_ssl = $::katello::candlepin_db_ssl,
+  Boolean $db_ssl_verify = $::katello::candlepin_db_ssl_verify,
+  Boolean $manage_db = $::katello::candlepin_manage_db,
+  String $qpid_hostname = $::katello::qpid_hostname,
+) {
+  include ::certs::qpid
```
timogoebel: Actually, candlepin itself just needs `::certs::candlepin`. The candlepin puppet module needs `qpid_client`. As we include that now, we can drop this line.

manifests/candlepin.pp
```diff
@@ -0,0 +1,54 @@
...
+    db_user                      => $db_user,
+    db_password                  => $db_password,
+    db_ssl                       => $db_ssl,
+    db_ssl_verify                => $db_ssl_verify,
+    manage_db                    => $manage_db,
+    subscribe                    => Class['certs::qpid', 'certs::candlepin'],
+  }
+
+  # TODO: Is this still needed with proper containment?
+  Class['certs::candlepin'] ~> Service['tomcat']
```
timogoebel: We should check this with a test. This has to be handled inside the candlepin module.

Gemfile
```diff
@@ -4,6 +4,7 @@
 source 'https://rubygems.org'
 
 gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '>= 4.6'
+gem 'facter', ENV.key?('FACTER_VERSION') ? "~> #{ENV['FACTER_VERSION']}" : '~> 2.4.6'
```
ekohl: Note to self: this is no longer needed

manifests/candlepin.pp
ekohl: Yes, there is a test for this but currently ```candlepin``` doesn't contain the classes so until we refactor that we need this. Maybe ```Class['candlepin::service']``` would be cleaner but let's leave that for later.

manifests/candlepin.pp
ekohl: Actually we use the variables from this class (```$qpid_ssl_{cert,key}```) and subscribe to it so I think it's better to include it here as well. In fact, I'm going to add ```include ::certs``` because we use those vars as well.

manifests/init.pp
ekohl: Would it be a problem if we only exposed the hostname variables and not the URLs for now?

ekohl: This is now implemented so you can no longer directly change the URLs. I'm not sure if this is the best solution.

ekohl:
Rebased to fix a merge conflict. I also re-ordered the changes into logical commits.

manifests/qpid.pp
```diff
@@ -1,34 +1,47 @@
...
     command  => "add queue ${candlepin_event_queue} --durable",
     unless   => "queues ${candlepin_event_queue}",
-    ssl_cert => $client_cert,
-    ssl_key  => $client_key,
+    ssl_cert => $::certs::qpid::client_cert,
+    ssl_key  => $::certs::qpid::client_key,
   } ->
   qpid::config::bind { ['entitlement.created', 'entitlement.deleted', 'pool.created', 'pool.deleted', 'compliance.created']:
     queue    => $candlepin_event_queue,
     exchange => $candlepin_qpid_exchange,
```
ekohl: https://github.com/Katello/puppet-qpid/pull/71 ensures that the chaining is there automatically. When that's in we can also change the above ```config_cmd``` to a queue type.

manifests/application.pp
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  include ::foreman
+  include ::foreman::plugin::tasks
+
+  Class['certs::apache'] ~> Class['foreman::service']
+
+  # Katello database seeding needs candlepin
+  Exec <| tag == 'cpinit' |> ->
+  foreman::plugin { 'katello':
```
ekohl: This is leading of a dependency cycle (as I was afraid of):

(Exec[foreman-rake-apipie:cache:index] => Foreman::Rake[apipie:cache:index] => Class[Foreman::Database] => Foreman::Plugin[katello] => Package[katello] => Foreman::Plugin[katello] => Exec[foreman-rake-db:seed] => Foreman::Rake[db:seed] => Class[Foreman::Database])

Maybe we need to revisit the chaining in puppet-foreman to first install plugins, then do the database.

manifests/candlepin.pp
ekohl: I believe https://github.com/Katello/puppet-candlepin/pull/81 should remove the need for it. Once it's merged we can remove this line and see if the tests still pass.

manifests/candlepin.pp
```diff
@@ -0,0 +1,55 @@
+# Katello configuration for pulp
```
ehelms: typo s/pulp/candlepin -- should this class get docs for parameters?

ekohl: I'll fix the typo, not sure about docs because I consider it a private class for now. The parameters are mostly there so you can unit test it without including katello (reducing test time)

manifests/init.pp
```diff
@@ -75,6 +75,24 @@
...
 #
+# $enable_candlepin::   Whether to enable candlepin
+#
+# $enable_qpid::        Whether to enable qpid
+#
+# $enable_qpid_client:: Whether to enable qpid client
+#
+# $enable_pulp::        Whether to enable pulp
+#
+# $enable_application:: Whether to enable application (katello web ui)
```
ehelms: Not sure I like the `enable` vocabulary, these are not optional, just how you manage them is. So I'd lean towards more of the `manage_` paradigm like we use elsewhere.

manifests/init.pp
```diff
@@ -84,7 +102,7 @@
   String $oauth_secret = $::katello::params::oauth_secret,
 
   String $post_sync_token = $::katello::params::post_sync_token,
-  Integer[0, 5000] $qpid_session_unacked = $::katello::params::qpid_session_unacked,
+  Optional[Integer[0, 5000]] $qpid_session_unacked = $::katello::params::qpid_session_unacked,
```
ehelms: Needs removing after recent updates

manifests/init.pp
```diff
@@ -115,104 +133,65 @@
...
-  class { '::qpid::client':
-    ssl                    => true,
-    ssl_cert_name          => 'broker',
-    ssl_cert_db            => $certs::nss_db_dir,
-    ssl_cert_password_file => $certs::qpid::nss_db_password_file,
-  } ~>
-  class { '::katello::qpid':
-    client_cert => $certs::qpid::client_cert,
-    client_key  => $certs::qpid::client_key,
+  $candlepin_url = "https://${candlepin_hostname}:8443/candlepin"
```
ehelms: Should port be configurable here?

manifests/init.pp
```diff
@@ -115,104 +133,65 @@
...
+  $pulp_url = "https://${pulp_hostname}/pulp/api/v2/"
```
ehelms: Should port be configurable here?

manifests/init.pp
```diff
@@ -115,104 +133,65 @@
...
+    Class['certs'] ~> Class['katello::qpid_client']
+  }
+
+  if $enable_pulp {
+    include ::katello::pulp
+    Class['certs'] ~> Class['katello::pulp']
   }
 
-  class { '::certs::foreman': }
+  if $enable_application {
```
ehelms: In what cases would this module not want to enable application? I feel like thats the major role of this puppet module.

manifests/pulp.pp
```diff
@@ -0,0 +1,61 @@
+# Katello configuration for pulp
+class katello::pulp {
+  include ::certs::qpid_client
+  # Because we re-use the CRL file
+  include ::katello::candlepin
```
ehelms: Will this lock Pulp to always requiring Candlepin on the same box?

manifests/qpid.pp
```diff
@@ -1,34 +1,47 @@
...
-    ssl_cert => $client_cert,
-    ssl_key  => $client_key,
+    ssl_cert => $::certs::qpid::client_cert,
+    ssl_key  => $::certs::qpid::client_key,
```
ehelms: May want to consider if we want to lock this in to the `certs` requirement or have this as configurable parameter

ehelms:
I don't entirely get the `application` manifest that was introduced, the rest make sense though.

manifests/init.pp
ekohl: I'm not sure what's best. I think for candlepin we could only expose the URL parameter rather than the hostname. @Klaas- what do you think?

manifests/init.pp
ekohl: Same as the candlepin URL

manifests/init.pp
ekohl: In case you want to deploy a pulp or candlepin on server A but the application on server B. Possibly you have multiple server Bs behind a load balancer.

manifests/pulp.pp
ekohl: It does look that way. I haven't looked into having a different method for the CRL file and suggestions are highly welcome.

ekohl:
Acceptance tests still fail because it needs an unreleased puppet-certs and right now there are still some other issues to fix which I want to focus on first.

ekohl:
Marking this WIP again since the whole pulp situation is still not very good.

A short status update might be good.

Some issues:
* Currently ```pulp.pp``` uses ```foreman::config::passenger::fragment``` which makes no sense if there's no Foreman vhost.
* The ```pulp_export_dir``` is part of ```pulp.``` but according to https://github.com/Katello/katello/blob/6aca54157e579d6312a9e4a49609df0ffa1685b1/app/lib/actions/katello/repository/export.rb#L15-L17 it should be on the Katello host. This assumption will fail if the two are on separate hosts. Should we create them on both hosts for now?

My first goal will be to to run 3 hosts:
* services
  * candlepin
  * pulp
  * qpid
* app
  * foreman + its plugins
* proxy
  * foreman-proxy

Once I can run that, I'll see about removing WIP again. I might try to rework the PR to split the classes but not introduce the enable_* parameters yet since I do think they improve the module. I also agree with @ehelms that manage_* would be a better name.

ekohl:
I've opened https://github.com/Katello/puppet-katello/pull/208 as a version 4 of the split classes. That version omits the enable_* parameters and improves test coverage, containment and dependencies. Right now I'm running that through the pipeline and I'd like to focus on getting that in first.

timogoebel:
> This assumption will fail if the two are on separate hosts. Should we create them on both hosts for now?

I'd suggest to just create the directory on the pulp host. If a user decides to split up the services and doesn't use the export feature, he will be fine. If he tries to export a repository, this will fail with a good error message.
My suggestion is to introduce the "enable_"-parameters regardless of that issue. We should file a Redmine under Katello for this and fix it there. My gut feeling is, that the copy task should be offloaded to a smart proxy (e.g. smart_proxy_pulp). In addition, a ticket with pulp should be filed so that you can specify a parameter to where pulp exports the files.

ekohl:
Replaced by https://github.com/Katello/puppet-katello/pull/209.

PR #209, opened 2017-08-29
[WIP] Fixes #20851 - Introduce manage flags

ekohl:
This allows managing only parts of the application. It is only a partial implementation since there are still dependencies left, but it hard fails when the user attempts to use a broken setup.

The issue with pulp is that it relies on adding fragments using the ```foreman::config::passenger::fragment```. I'm looking into using concat fragments and when that's done, I think it can be merged. https://github.com/puppetlabs/puppetlabs-apache/pull/1659 might make this easier in the future.

I'm not sure yet how to handle the candlepin CRL so suggestions welcome.


timogoebel:
@ekohl: Sorry for the delay in taking a look at your PRs.

Regarding the CRL issue: I think the only way to handle that correctly is setting up an external sync service. The candlepin host has to publish the CRL via http and pulp (the client if you want) has to download the CRL regularly. The poor man's solution would probably be a "cron & wget".
But I'm not sure we want to support that via the puppet modules.

manifests/pulp.pp
```diff
@@ -15,6 +15,7 @@
   Stdlib::Absolutepath $repo_export_dir = $::katello::repo_export_dir,
   String $repo_export_dir_owner = $::katello::user,
   String $repo_export_dir_group = $::katello::group,
+  Boolean $manage_httpd = $::katello::pulp_manage_httpd,
```
ekohl: The reasoning behind this parameter is that we have 3 ways of deploying pulp:
* Deploy as part of the foreman vhost (puppet-katello)
* Deploy as part of the devel vhost (puppet-katello_devel)
* Deploy a full vhost (puppet-foreman_proxy_content)

They are all almost identical and I'd like a solution where we can reuse. At least katello & katello_devel should be doable.

manifests/pulp.pp
```diff
@@ -57,9 +58,18 @@
...
-  foreman::config::passenger::fragment { 'pulp':
-    content     => file('katello/pulp-apache.conf'),
-    ssl_content => file('katello/pulp-apache-ssl.conf'),
+  unless $manage_httpd {
+    foreman::config::passenger::fragment { 'pulp':
+      content     => file('katello/pulp-apache.conf'),
+      ssl_content => file('katello/pulp-apache-ssl.conf'),
+    }
+
+    file { $repo_export_dir:
```
sean797: Why move this? I think people export content regardless of Apache. Some people export onto external media and import into another Katello server, I know others who export it so its available via HTTP and change sync across the network. https://theforeman.org/plugins/katello/3.4/user_guide/disconnected/index.html

manifests/init.pp
```diff
@@ -69,6 +69,20 @@
 #
 # $candlepin_manage_db:: Boolean indicating whether a database should be installed, this includes db creation and user
 #
+# $manage_candlepin::   Whether to manage candlepin
+#
+# $manage_qpid::        Whether to manage qpid
+#
+# $manage_pulp::        Whether to manage pulp. Currently requires $manage_application and $manage_candlepin to be true as well.
+#
+# $manage_application:: Whether to manage application (katello web ui)
```
sean797: I don't really like the name of this, how about `manage_katello` as that fits with the other params naming convention like `manage_candlepin` ect.. ?

ekohl:
> So this allows our RoR app on one node, and Pulp, qpidd, candlepin on another?

That is the goal yes. You're right that it will break things and we just don't know what yet.

manifests/pulp.pp
ekohl: This is an ugly hack that allowed me to move forward. The problem is that it tries to create this directory owned by ```foreman:foreman``` which doesn't exist if you only install pulp on a node. See https://groups.google.com/forum/#!topic/foreman-dev/sofiVxnbLGQ for some background.

manifests/init.pp
ekohl: Naming is hard and I'm fully open to other names. The reason I chose application was that this module is also named katello so ```$katello::manage_katello``` felt odd. UI can't be used because it also includes other aspects of it (tasks/dynflow daemon). That said, I'm not attached to ```manage_application```.

timogoebel:
> You're right that it will break things and we just don't know what yet.

Actually, not that much. The main things worked quite fine.
I'm planning on picking this up soon. I'll let you know if I find something. A tracker issue for redmine sound like a super idea @sean797. Would you mind setting one up?
One issue is worth a Redmine is definitely this: https://github.com/Katello/katello/blob/6aca54157e579d6312a9e4a49609df0ffa1685b1/app/lib/actions/katello/repository/export.rb#L15-L17

ekohl:
> Actually, not that much. The main things worked quite fine.

It'll be subtle things and I do agree with @sean797 that they shouldn't be a blocker. We can always release it as experimental and warn users in the release notes. A redmine tracker is a good thing and I can create some issues.

ekohl:
See http://projects.theforeman.org/issues/20850 and the related issues.

manifests/init.pp
sean797: Yea, agreed i dont like `$katello::manage_katello`. I'll think on it.

manifests/init.pp
```diff
@@ -106,15 +120,49 @@
...
+    include ::katello::candlepin
+  }
+
+  if $manage_qpid {
+    include ::katello::qpid
+  }
+
+  if $manage_pulp {
+    unless $manage_candlepin {
+      fail('Pulp relies on candlepin for the CRL. Set $manage_candlepin to true.')
```
ehelms: Hrmm, we should look more into this because we don't enforce or use CRL.

ekohl: I was thinking of a switch to use local or a simple script to retrieve the CRL and reload apache if it changed but first I want to get the setup with 2 servers running.

timogoebel: I'd really appreciate it, if this check would be removed. There might be users (me :-P) who want to ignore this dependency for starters.

ekohl: I hadn't arrived there yet and wanted to be safe. My thinking was a parameter like ```Enum['disabled', 'candlepin'] $pulp_crl``` that will allow us to add a script later that retrieves the CRL. Thoughts?

timogoebel: Yep, this sounds like a good compromise. I like the idea.

ekohl:
Includes https://github.com/Katello/puppet-katello/pull/218 with a minor update so it uses ```manage_application``` to determine the default.

ekohl:
Rebased. I think this is ready for review.

timogoebel:
@ekohl: ... you might want to remove the "WIP" in the title, then. :-)

ekohl:
@timogoebel when I ran this on a pipeline the proxy had a timeout starting mongodb. Not sure if that was a one time thing or reproduceable. Going to retry to find that out. Until then the WIP sort of makes sense :)

manifests/init.pp
```diff
@@ -73,6 +73,20 @@
 #
 # $candlepin_manage_db:: Boolean indicating whether a database should be installed, this includes db creation and user
 #
+# $manage_candlepin::   Whether to manage candlepin
+#
+# $manage_qpid::        Whether to manage qpid
+#
+# $manage_pulp::        Whether to manage pulp. Currently requires $manage_application and $manage_candlepin to be true as well.
```
timogoebel: > `Currently requires $manage_application and $manage_candlepin to be true as well.``

Is that true? Why is that?

manifests/pulp.pp
```diff
@@ -49,14 +62,28 @@
...
+  } else {
+    foreman::config::passenger::fragment { 'pulp':
+      content     => file('katello/pulp-apache.conf'),
+      ssl_content => file('katello/pulp-apache-ssl.conf'),
+    }
+
+    foreman::config::passenger::fragment { 'pub':
+      content     => file('katello/pub-apache.conf'),
+      ssl_content => file('katello/pub-apache.conf'),
+    }
```
timogoebel: Do we also need to proxy GPG to the foreman host?

manifests/init.pp
ekohl: Leftover. See, comments always go out of date since there are no tests for it.

manifests/pulp.pp
ekohl: I don't think so because candlepin runs on the host itself so there's no need to proxy it. If they are split, we might need to figure out what's correct. If a smart proxy host proxies it to an application host which proxies it to a candlepin host that's somewhat complexer than it might need to be.

manifests/init.pp
```diff
@@ -79,6 +79,20 @@
 #                        mongo database has slow I/O, then setting a higher number may resolve issues where workers are
 #                        going missing incorrectly.
 #
+# $manage_candlepin::    Whether to manage candlepin
+#
+# $manage_qpid::         Whether to manage qpid
+#
+# $manage_pulp::         Whether to manage pulp. Currently requires $manage_application and $manage_candlepin to be true as well.
+#
+# $manage_application::  Whether to manage application (katello web ui)
```
ehelms: I think I keep getting hung up on application and wonder if something somewhat more specified could work. Ideas that come to mind: server, web_server, rails_server, foreman_server, rails_application, foreman_application

manifests/init.pp
```diff
@@ -120,15 +136,41 @@
...
+  if $manage_candlepin {
+    include ::katello::candlepin
+  }
+
+  if $manage_qpid {
+    include ::katello::qpid
+  }
+
+  if $manage_pulp {
+    include ::katello::repo
```
ehelms: This can be dropped I think given we are using Pulp's repositories directly now.

manifests/pulp.pp
ehelms: This will cause problems though -- https://github.com/theforeman/puppet-pulp/commit/5ff7611c6e577b95f5a3f096830c4e3097c4db2d

manifests/init.pp
ekohl: I like foreman_application

manifests/init.pp
ekohl: Should we use pass ```manage_repo``` to the pulp class to get that repository installed?

ekohl:
Updated

manifests/pulp.pp
```diff
@@ -46,15 +59,39 @@
...
     subscribe              => Class['certs', 'certs::qpid_client'],
     worker_timeout         => $pulp_worker_timeout,
   }
 
   contain ::pulp
 
-  foreman::config::passenger::fragment { 'pulp':
-    content     => file('katello/pulp-apache.conf'),
-    ssl_content => file('katello/pulp-apache-ssl.conf'),
+  if $manage_httpd {
```
ekohl: This is still not really great. The third use case is katello devel which has a third vhost. I think a better solution would be to directly use apache vhost concat fragments and just pass in the title of the vhost.

ekohl:
Rebased, hopefully I didn't mess things up while trying to fix the conflicts.

ekohl:
Closing due to age.

PR #211, opened 2017-09-05
Fixes #20857 - override post_sync_url

sean797:


ekohl:
What do you think about using ```$::foreman::foreman_url```? An alternative is re-using ```$::foreman::servername```. It's messy we have both and I'd prefer ```$::foreman::foreman_url```. The benefit is that when you'd include ```foreman::cli``` on the same host you automatically configure that correct (though I don't think katello-installer exposes that option).

sean797:
@ekohl Yea, nice idea. Updated to use `$::foreman::foreman_url` Thanks!

manifests/application.pp
```diff
@@ -17,11 +17,13 @@
   Optional[String] $proxy_username = $::katello::proxy_username,
   Optional[String] $proxy_password = $::katello::proxy_password,
 ) {
+  include ::foreman
   include ::certs
   include ::certs::apache
   include ::certs::foreman
   include ::certs::pulp_client
 
+  $application_host = $::foreman::foreman_url
```
ekohl: It might simplify the config template if you define ```$post_sync_url``` here.

sean797:
Hows that @ekohl ?

manifests/application.pp
```diff
@@ -17,11 +17,14 @@
...
   Optional[String] $proxy_password = $::katello::proxy_password,
 ) {
+  include ::foreman
   include ::certs
   include ::certs::apache
   include ::certs::foreman
   include ::certs::pulp_client
 
+  $post_sync_url = "${$::foreman::foreman_url}${deployment_url}/api/v2/repositories/sync_complete?token=${post_sync_token}"
+  $application_host = $::foreman::foreman_url
```
ekohl: Is this used anywhere?

ekohl:
This looks a lot like a change I already wanted to make but hadn't gotten around to :)

sean797:
ah, damn! That was left over from before. Thanks @ekohl

manifests/application.pp
```diff
@@ -17,11 +17,13 @@
   Optional[String] $proxy_username = $::katello::proxy_username,
   Optional[String] $proxy_password = $::katello::proxy_password,
 ) {
+  include ::foreman
   include ::certs
   include ::certs::apache
   include ::certs::foreman
   include ::certs::pulp_client
 
+  $post_sync_url = "${$::foreman::foreman_url}${deployment_url}/api/v2/repositories/sync_complete?token=${post_sync_token}"
```
ekohl: I'm surprised this works. Usually we have ```${::foreman::foreman_url}```. The tests appear to pass so this is valid syntax?

sean797: Interesting, it must be. I was actually a typo so I've update to match the rest of our code base.

ekohl:
Thanks! This removes an item from my (mental) list of things to fix.

PR #212, opened 2017-09-08
Fixes #20879 - Use a different oauth for Pulp & Candlepin

sean797:


spec/classes/katello_pulp_spec.rb
```diff
@@ -11,7 +11,8 @@
         let :pre_condition do
           <<-EOS
           class { '::katello':
-            oauth_secret     => 'secret',
+            pulp_oauth_key        => 'key',
```
ekohl: nit: the alignment is off here.

sean797:
@ekohl thanks for taking a look. I'm fairly certain this wont break any upgrades as puppet will update the keys & secrets everywhere and restart services. But I haven't tested an upgrade.

ekohl:
> @ekohl thanks for taking a look. I'm fairly certain this wont break any upgrades as puppet will update the keys & secrets everywhere and restart services. But I haven't tested an upgrade.

Nowadays you're right it should work because the installer ignores unknown answers and users are not expected to override this but it's still a major version bump per semver.

sean797:
Raised https://github.com/Katello/puppet-katello/pull/212/files for the migration

sean797:
Raised https://github.com/Katello/puppet-katello/pull/212 for the migration

ekohl:
Looks like I caused a merge conflict. Other than that I'd like @ehelms to have a look at this and the migration before I merge it.

sean797:
Updated, thanks!

ehelms:
I wonder if we shouldn't just drop the oauth here for Pulp. We do not use it anymore, as we use client certificate for communication. Oauth is also a deprecated auth for Pulp. That would simplify this update.

ekohl:
[split candlepin/pulp oauth](https://trello.com/c/afmNVNIX/513-split-candlepin-pulp-oauth)

sean797:
@ehelms I didn't know that, I'll look into remove Katello's support for it 👍

sean797:
Closed in favor of https://github.com/Katello/puppet-katello/pull/223

PR #213, opened 2017-09-08
Ensure the candlepin keystore symlink

ekohl:
This was previously in puppet-certs as /usr/share/tomcat/conf/keystore
but it doesn't belong there. Because /usr/share/tomcat/conf is a symlink
to /etc/tomcat we can safely manage it here as well without duplicating
resources.

### DIFF
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -51,4 +51,13 @@ class katello::candlepin (
   }
 
   contain ::candlepin
+
+  file { '/etc/tomcat/keystore':
+    ensure  => link,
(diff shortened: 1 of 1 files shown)


ekohl:
Original code: https://github.com/Katello/puppet-certs/blob/fe9737637bd2c484e7606d1b158f3d31ec65f113/manifests/candlepin.pp#L100-L105

ekohl:
Currently we rely on the katello meta package to install things but I added the ordering in case we don't.

ekohl:
Also made it rely on ```certs::candlepin```. Probably not really needed but better to be sure.

PR #215, opened 2017-09-24
candlepin: use own certs for qpid

timogoebel:
The `qpid_ssl_cert` and `qpid_ssl_key` parameters for `candlepin` are just to [set up an exchange](https://github.com/Katello/puppet-candlepin/blob/bdcfccd827e2f8f5477e37a1f20923641fb4439a/manifests/qpid.pp#L7-L10).

This can be done with candlepin's qpid certificates and removed the dependency for qpid's certificates.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -22,6 +22,8 @@ class katello::application (
   include ::certs::apache
   include ::certs::foreman
   include ::certs::pulp_client
+  include ::certs::qpid
+  include ::katello::qpid_client
 
   $post_sync_url = "${::foreman::foreman_url}${deployment_url}/api/v2/repositories/sync_complete?token=${post_sync_token}"
   $candlepin_ca_cert = $::certs::ca_cert
@@ -50,6 +52,7 @@ class katello::application (
   include ::foreman::plugin::tasks
 
   Class['certs', 'certs::ca', 'certs::apache'] ~> Class['apache::service']
+  Class['certs', 'certs::ca', 'certs::qpid'] ~> Class['foreman::plugin::tasks']
 
   # Katello database seeding needs candlepin
   package { $package_names:
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -14,9 +14,7 @@ class katello::candlepin (
   String $qpid_hostname = $::katello::qpid_hostname,
 ) {
   include ::certs
-  include ::certs::qpid
   include ::certs::candlepin
-  include ::katello::qpid_client
 
   class { '::candlepin':
     user_groups                  => $user_groups,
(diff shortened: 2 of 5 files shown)


timogoebel:
Tests currently fail, because `::katello::qpid_client` isn't in the catalog anymore. I'm curious, if `::katello::qpid_client` is just needed by something else or it's just unneeded clutter.

timogoebel:
I have a feeling, that actually the katello rails application needs to have `include ::katello::qpid_client`. I'll continue looking into this and update the PR accordingly.

timogoebel:
Yeah, my expectation was correct: The Katello rails app needs `katello::qpid_client` to work properly. You can see this in `/var/log/foreman/dynflow_executor.output`:

```
Starting dynflow with the following options: {:foreman_root=>"/usr/share/foreman", :process_name=>"dynflow_executor", :pid_dir=>"/usr/share/foreman/tmp/pids", :log_dir=>"/usr/share/foreman/log", :wait_attempts=>300, :wait_sleep=>1, :executors_count=>1, :memory_limit=>0, :memory_init_delay=>7200, :memory_polling_interval=>60}
Everything ready for world: 42678733-a5ff-4ee1-a860-f2dca3bec4bd
terminate called after throwing an instance of 'qpid::types::Exception'
  what():  Failed to initialise SSL: Failed: NSS error [-8015] (/builddir/build/BUILD/qpid-cpp-1.36.0/src/qpid/sys/ssl/util.cpp:100) (/builddir/build/BUILD/qpid-cpp-1.36.0/src/qpid/client/SslConnector.cpp:149)
```

timogoebel:
This is still missing a notify from `Class['certs::qpid'] ~> Service['foreman-tasks']`.

manifests/application.pp
```diff
@@ -22,6 +22,8 @@
   include ::certs::apache
   include ::certs::foreman
   include ::certs::pulp_client
+  include ::certs::qpid
```
ekohl: Shouldn't qpid_client be sufficient? Or can we make it sufficient?

timogoebel: Well, `certs::qpid_client` just rolls out certificates for the qpid-config tool. `certs::qpid` rolls out the whole nss certstore. We need the latter for the app.
`::katello::qpid_client` does include `include ::certs::qpid`, so this is here just for completeness' sake.

ekohl: If we're not using it in this class then I'd consider it an implementation detail of ```qpid_client```, but it might even be better to just include ```qpid::client``` in this class rather than split it off if there's just 1 use of it.

ekohl:
> This is still missing a notify from Class['certs::qpid'] ~> Service['foreman-tasks'].

With https://github.com/theforeman/puppet-foreman/pull/530 that'll be nicely wrapped in ```foreman::service```

timogoebel:
@ekohl: Added the notify and a test for it so it works with the current implementation of `puppet-foreman`.

manifests/application.pp
timogoebel: I left this in now, as it's required for the notify of foreman-tasks.

ekohl:
Thanks!

PR #216, opened 2017-09-24
qpidd: hostname for config cmds

timogoebel:
This PR allows configuring the hostname that is used to connect to `qpidd` while setting up queues etc. This is necessary when a developer sets `interface` to something else than `lo`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -41,6 +41,10 @@
 #
 # $qpid_wcache_page_size:: The size (in KB) of the pages in the write page cache
 #
+# $qpid_interface::     The interface qpidd listens to.
+#
+# $qpid_hostname::      Hostname used to connect to qpidd.
+#
 # $package_names::      Packages that this module ensures are present instead of the default
 #
 # $manage_repo::        Whether to manage the yum repository
@@ -79,6 +83,8 @@ class katello (
 
   String $post_sync_token = $::katello::params::post_sync_token,
   Integer[0, 1000] $qpid_wcache_page_size = $::katello::params::qpid_wcache_page_size,
+  String $qpid_interface = $::katello::params::qpid_interface,
+  String $qpid_hostname = $::katello::params::qpid_hostname,
   Integer[1] $num_pulp_workers = $::katello::params::num_pulp_workers,
   Optional[Integer] $max_tasks_per_pulp_worker = $::katello::params::max_tasks_per_pulp_worker,
   Optional[Stdlib::HTTPUrl] $proxy_url = $::katello::params::proxy_url,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -79,6 +83,8 @@
 
   String $post_sync_token = $::katello::params::post_sync_token,
   Integer[0, 1000] $qpid_wcache_page_size = $::katello::params::qpid_wcache_page_size,
+  Optional[String] $qpid_interface = $::katello::params::qpid_interface,
```
ekohl: This and ```qpid_hostname``` can just be a ```String``` since we have a default.

manifests/qpid.pp
```diff
@@ -4,7 +4,8 @@
   String $candlepin_event_queue = $::katello::candlepin_event_queue,
   String $candlepin_qpid_exchange = $::katello::candlepin_qpid_exchange,
   Integer[0, 5000] $wcache_page_size = $::katello::qpid_wcache_page_size,
-  String $interface = 'lo',
+  String $interface = $::katello::qpid_interface,
+  Optional[String] $hostname = $::katello::qpid_hostname,
```
ekohl: This can also be a ```String```

timogoebel:
@ekohl: Thanks. Done.

ehelms:
IIRC, we originally locked this to localhost due to security concerns. I think those have since been fixed, but I would need to dig back in the past. Perhaps @stbenjam remembers.

timogoebel:
> IIRC, we originally locked this to localhost due to security concerns.

I think, this should not be a problem at all. The parameters default to localhost (which adds just another layer of security that relies heavily of the security of your host) and are listed as "advanced parameters".  The main security should come through the TLS connection. I think, there is no need to sorry.

stbenjam:
Yea, in general we don't want to expose the broker to the outside world.  Content hosts have valid client SSL certificates and can connect to the broker (which is fine, they used to do that before we had the dispatch router), but they can also see the pulp queues. 

Theoretically, if one found a bug in pulp it could be exploited by a compromised content host injecting a message into the queue.  The dispatch router is more locked down as it only allows communication in certain directions, it's not possible for a client to inject a message into the various queues used by pulp.

If qpidd is going to run on a different host than the Katello, ideally we'd need to allow access only to the  Katello box by iptables.

timogoebel:
@stbenjam: Thanks for the clarification. That does make sense. For our use case, the connections are strictly limited by our SDN solution. So we're covered there.
Do you guys think, we should extend the parameter description with a warning that this could be an issue? Or can we merge this without?

stbenjam:
I'm not sure, but I guess since it's already under advanced that is probably enough.

PR #217, opened 2017-10-06
Introduce katello::postgresql

ekohl:
Currently it's impossible to tune postgresql from the installer. This patch allows setting tuning parameters through hiera while making no difference to the current installation.

In the future this class could be exposed as a top level manage flag so you can set up split deployments.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -25,6 +25,10 @@ class katello::application (
   include ::certs::qpid
   include ::katello::qpid_client
 
+  if $::foreman::db_manage {
+    include ::katello::postgresql
+  }
+
   $post_sync_url = "${::foreman::foreman_url}${deployment_url}/api/v2/repositories/sync_complete?token=${post_sync_token}"
   $candlepin_ca_cert = $::certs::ca_cert
   $pulp_ca_cert = $::certs::katello_server_ca_cert
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -16,6 +16,10 @@ class katello::candlepin (
   include ::certs
   include ::certs::candlepin
 
+  if $manage_db {
+    include ::katello::postgresql
+  }
+
   class { '::candlepin':
     user_groups                  => $user_groups,
     oauth_key                    => $oauth_key,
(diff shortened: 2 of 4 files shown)


sean797:
Would this be better done in `foreman::database::postgresql`? Then Foreman users can also take advantage of this.

I guess you'd have to duplicate it in `candlepin::database::postgresql` as well, but at least then Katello is getting more integrated as a proper Foreman plugin. What do you think?

ekohl:
That's one of the hard issues. I see puppet-katello more as a profile module but foreman-installer has no profile modules (though puppet-foreman_proxy has the proxydns and proxydhcp which could be considered profiles). We had a similar discussion with the postgresql HA setup and there's no proper place for it. It's sort of trying to put business logic in a generic solution.

The naming of things in the installer is something terribly difficult. Right now there's a very strong mapping to the puppet modules and patterns but I'm starting to have some doubts because we're running into limitations. Sadly I don't have a concrete solution.

sean797:
I guess another reason to create a `puppet-foreman_profile` module. For now I think this is the best we can do. ACK from me. Thanks!

ekohl:
@mmoll I'd like your thoughts on this given we want to eventually merge the whole katello-installer in the foreman installer.

mmoll:
As this is still contained to the katello module, this is perfectly fine.

ekohl:
I've taken a stab at solving it in puppetlabs-apache instead. Doesn't appear to work just yet but https://github.com/puppetlabs/puppetlabs-postgresql/pull/950

ehelms:
I read through the comments and I still don't understand why this is Katello specific. I get the profile arguments, but if this is about tuning Postgres for performance then that seems useful to all versions of Foreman running with Postgres.

ekohl:
> I read through the comments and I still don't understand why this is Katello specific. I get the profile arguments, but if this is about tuning Postgres for performance then that seems useful to all versions of Foreman running with Postgres.

That is exactly the problem I have too -- I couldn't find the right place to put it in any of our modules. That's why I ended up opening a PR in puppetlabs-postgresql which is IMHO still the preferred solution. https://github.com/puppetlabs/puppetlabs-postgresql/pull/950 has been updated and I think it's in a mergable shape.

ekohl:
Closing in the hope https://github.com/puppetlabs/puppetlabs-postgresql/pull/950 gets merged soon.

PR #218, opened 2017-10-09
pulp: manage_httpd

timogoebel:
This introduces a parameter `manage_httpd` that is required for setups where pulp should live on a standalone host.
This parameter is not exposed in init.pp on purpose for now. Users that want to use this can do something like this or use Hiera:

```puppet
class { '::katello':
  manage_candlepin   => false,
  manage_qpid        => false,
  manage_pulp        => false, # explicitly declared below to overwrite parameters
  manage_application => false,
  qpid_hostname      => 'qpid.example.com',
  proxy_url          => 'http://proxy.example.com',
  proxy_port         => 8000,
}

class { '::katello::pulp':
  manage_httpd => true,
}
```

This also removes one duplicate of the pubdir declaration.

### DIFF
--- a/manifests/pulp.pp
+++ b/manifests/pulp.pp
@@ -15,11 +15,24 @@ class katello::pulp (
   Stdlib::Absolutepath $repo_export_dir = $::katello::repo_export_dir,
   String $repo_export_dir_owner = $::katello::user,
   String $repo_export_dir_group = $::katello::group,
+  Boolean $manage_httpd = false,
 ) {
   include ::certs
   include ::certs::qpid_client
   # Because we re-use the CRL file
   include ::katello::candlepin
+
+  if $manage_httpd {
+    include ::certs::apache
+
+    $https_cert = $::certs::apache::apache_cert
+    $https_key  = $::certs::apache::apache_key
+    $ca_cert    = $::certs::katello_server_ca_cert
+  } else {
+    $https_cert = undef
+    $https_key  = undef
+    $ca_cert    = undef
+  }
 
   class { '::pulp':
     oauth_enabled          => true,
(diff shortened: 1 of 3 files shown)


manifests/pulp.pp
```diff
@@ -52,14 +65,28 @@
...
+    ca_cert                => $ca_cert,
     subscribe              => Class['certs', 'certs::qpid_client'],
   }
 
   contain ::pulp
 
-  foreman::config::passenger::fragment { 'pulp':
-    content     => file('katello/pulp-apache.conf'),
-    ssl_content => file('katello/pulp-apache-ssl.conf'),
+  if $manage_httpd {
```
timogoebel: I just noticed, this is missing `/pub` for the http (80) vhost.

ekohl: Could you have a look at this again?

sean797:
ping @timogoebel whats the status here? :-)

laugmanuel:
Any updates on this?
@ekohl Your commit seems to do what the PR was intended for.

timogoebel:
@laugmanuel: Do you mind taking this task over?

laugmanuel:
@timogoebel: No, not at all. But I don't think there is much left to do.

ekohl:
The majority of the work is around the certificate handling and testing it out end-to-end. I have a few other larger tasks but I still want this work to finish.

laugmanuel:
If there is a relatively fast way to setup a test environment, I can help test the module.
What I can say is that we use the version of Timo in our environment for some time now.

laugmanuel:
Given this MR is now over 2 years old and a lot of things changed in this Puppet module, I would suggest that I open a new MR based on the current version (rebase would result in changes to more or less every line in this MR ).

@ekohl, @timogoebel thoughts on this?

timogoebel:
@laugmanuel: Yes, please do. I'll close this PR for now.

PR #219, opened 2017-10-09
pulp: remove candlepin consumers crl

timogoebel:
This PR removes the dependency for candlepin from pulp.

This has never worked in the past. We need to properly design this and should implement a working solution.

https://github.com/Katello/puppet-pulp/blob/3bf701bacc8d98cdb49ac37d301322fbce941163/manifests/config.pp#L118-L124

The `refreshonly => true` requies the exec to be notified before it runs. It's never notified.

### DIFF
--- a/manifests/pulp.pp
+++ b/manifests/pulp.pp
@@ -18,8 +18,6 @@ class katello::pulp (
 ) {
   include ::certs
   include ::certs::qpid_client
-  # Because we re-use the CRL file
-  include ::katello::candlepin
 
   class { '::pulp':
     oauth_enabled          => true,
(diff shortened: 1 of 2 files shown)


ekohl:
Looks like we both forgot about the tests :)

timogoebel:
Somehow github doesn't update this PR to `https://github.com/timogoebel/puppet-katello/commit/b8e05188b23a58ce0df6f0d48096371176fa800b` :-(

ekohl:
@jlsherrill I think you confirmed this has never properly worked so we can remove it.

jlsherrill:
Correct, pulp no longer cares about the crl at all, and relies on apache to do it:  https://docs.pulpproject.org/user-guide/server.html#crl-support 

so puppet-pulp likey needs to be updated to support that.  From a katello perspective we've not felt it a priority to implement this properly, especially since we currently have no way to get these crls onto the smart proxies.  Its a solvable problem, just not one that we've cared to spend the resources on.  👍  to this!

ekohl:
Thanks @timogoebel for the patch and @jlsherrill for confirming.

PR #223, opened 2017-10-25
Fixes #21464 - Remove Pulp oauth support

sean797:
If we are happy with the renaming I will raise a PR with a migration to katello-installer.


ehelms:
@sean797 I appreciate you doing this, and I apologize up front for piling on. To make this complete, I think we need to in parallel remove the oauth support from Katello itself since this removes the ability to configure it. See:

https://github.com/Katello/katello/blob/master/app/services/katello/pulp/server.rb#L25 (and any place those configuration values are referenced)

ehelms:
Deleting in the migration and having them re-generated and thus re-populated across CP configuration and Katello configuration should be fine at this juncture.

sean797:
@ehelms I already created https://github.com/Katello/katello/pull/7031 :-)

Created https://github.com/Katello/katello-installer/pull/560 for the migration.

ekohl:
Since it defaults to false I think we can safely merge this even without the katello core being merged.

sean797:
Agreed @ekohl The Katello PR is now merged anyway.

manifests/params.pp
```diff
@@ -48,9 +48,8 @@
   $repo_export_dir = '/var/lib/pulp/katello-export'
 
   # OAUTH settings
-  $oauth_key = 'katello'
-  $oauth_token_file = 'katello_oauth_secret'
-  $oauth_secret = cache_data('foreman_cache_data', $oauth_token_file, random_password(32))
+  $candlepin_oauth_key = cache_data('foreman_cache_data', 'candlepin_oauth_key', random_password(32))
```
ehelms: Do we really want to set this to random? It is kinda nice that the key is set to `katello` as it clearly indicates who is using this to talk to CP.

ehelms:
@sean797 Do you mind completing this by cleaning up puppet-foreman_proxy_content of oauth references for Pulp as well?

manifests/params.pp
sean797: @ehelms The Foreman modules use a random string, though I do tend to agree especially as we move towards split deployments.

I'll change it.

https://github.com/theforeman/puppet-foreman/blob/master/manifests/params.pp#L220
https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/params.pp#L359

sean797:
ah @ehelms forgot about that... raised https://github.com/theforeman/puppet-foreman_proxy_content/pull/151

PR #225, opened 2017-11-03
Refs #15896 - Update bind address

chris1984:


manifests/params.pp
```diff
@@ -55,7 +55,7 @@
   $post_sync_token = cache_data('foreman_cache_data', 'post_sync_token', random_password(32))
 
   # Subsystems settings
-  $candlepin_url = "https://${::fqdn}:8443/candlepin"
+  $candlepin_url = 'https://127.0.0.1:8443/candlepin'
```
ehelms: I think we want to prefer `localhost` here instead of the IP address for ease of reference and certificates?

sean797: The users firewall should limit access to tomcat. Is there any benefit in changing the default?

ekohl:
There's a merge conflict now. Added the Backwards incompatible label since it currently is.

PR #231, opened 2018-01-22
Refs #22338 - add worker_timeout param

chris1984:


manifests/init.pp
```diff
@@ -75,6 +75,10 @@
 #
 # $rest_client_timeout:: Timeout for Katello rest API
 #
+# $pulp_worker_timeout:: The amount of time (in seconds) before considering a worker as missing. If Pulp's
+#                        mongo database has slow I/O, then setting a higher number may resolve issues where workers are
+#                        going missing incorrectly. Defaults to 30.
```
ekohl: The default in params.pp is 60. Not sure which we should default to but they should match :)

manifests/init.pp
```diff
@@ -89,6 +93,7 @@
   String $qpid_interface = $::katello::params::qpid_interface,
   String $qpid_hostname = $::katello::params::qpid_hostname,
   Integer[1] $num_pulp_workers = $::katello::params::num_pulp_workers,
+  Integer[0] $pulp_worker_timeout = $katello::params::pulp_worker_timeout,
```
sean797: Do you mind also properly namespacing this 

```
$::katello::params::pulp_worker_timeout,

chris1984:
@ekohl can I get another review please?

chris1984:
@ekohl can you review again when you get a moment please, tests are failing for a different reason.

ekohl:
Looks like we need to pin rdoc in modulesync :(

PR #237, opened 2018-02-26
Use configuration to determine content types

jlsherrill:


jlsherrill:
Corresponds with this PR: https://github.com/Katello/katello/pull/7204
Opening a a FPC Pr next and packaging pr to remove -ostree rpm

spec/classes/katello_application_spec.rb
```diff
@@ -9,7 +9,11 @@
         {
           :package_names          => ['tfm-rubygem-katello'],
           :enable_ostree          => false,
-          :rubygem_katello_ostree => 'tfm-rubygem-katello_ostree',
+          :enable_yum             => true,
+          :enable_file            => true,
+	  :enable_puppet          => true,
```
sean797: This passed linting ? :thinking:

manifests/pulp.pp
```diff
@@ -35,10 +40,10 @@
     manage_httpd           => false,
     manage_plugins_httpd   => true,
     manage_squid           => true,
-    enable_rpm             => true,
-    enable_deb             => true,
-    enable_puppet          => true,
-    enable_docker          => true,
+    enable_rpm             => $enable_yum or $enable_file,
```
sean797: same as foreman_proxy_content.. please use `enable_iso` param

spec/classes/katello_application_spec.rb
sean797: Its because you used tabs, you cowboy :cowboy_hat_face:

jlsherrill: errr what???  /me checks vim configuration

jlsherrill: :cry: newer-ish box didn't have my normal config

jlsherrill: We apologies for the tabs, those responsible have now been sacked.

manifests/application.pp
```diff
@@ -65,14 +70,7 @@
...
     content => template('katello/katello.yaml.erb'),
-    notify  => [Class['foreman::service', 'foreman::plugin::tasks'], Foreman::Rake['db:seed']],
-  }
-
-  if $enable_ostree {
-    package { $rubygem_katello_ostree:
-      ensure => installed,
-      notify => [Class['foreman::service', 'foreman::plugin::tasks'], Foreman::Rake['apipie:cache:index']],
-    }
+    notify  => [Class['foreman::service', 'foreman::plugin::tasks'], Foreman::Rake['db:seed'], Foreman::Rake['apipie:cache:index']],
```
ekohl: If you depend on theforeman-foreman >= 9.0.0 then you can leave out the tasks notification since the module handles that itself.

jlsherrill:
@ekohl what makes this backwards incompatible?

ekohl:
> @ekohl what makes this backwards incompatible?

My misreading. I thought ```application.pp``` was ```init.pp``` and that would be backwards incompatible.

jlsherrill:
@ekohl ahhh good point :+1:

ekohl:
@jlsherrill should we consider this backwards incompatible then?

jlsherrill:
@sean797 you good with this too?

sean797:
Yes - haven’t tested, but looks good!

jlsherrill:
@ekohl i guess it depends on how you define backwards incompatible.  Its compatible as far as you can continue to use it with no param or class changes and you will get the same result.  Its incompatible in that this change won't work on older versions of katello and still maintain the 'enable_ostree' param's purpose.

ekohl:
@jlsherrill in that case I consider it backwards incompatible. That's no big deal because I want to land the split deployments in the next release anyway, even if the certificate support isn't all there yet.

manifests/application.pp
jlsherrill: oh i missed this comment, let me do that.

PR #238, opened 2018-03-01
Refs #22513 - Add qpid tuning to katello

chris1984:


manifests/init.pp
```diff
@@ -89,6 +89,10 @@
 #                        mongo database has slow I/O, then setting a higher number may resolve issues where workers are
 #                        going missing incorrectly.
 #
+# $qpid_mgmt_pub_interval:: Controls the interval at which the broker sends
+#                           updated information (stats, etc.) to the management console
+# $qpid_open_file_limit:: Limit number of open files - systemd distros only
```
ehelms: These are fairly advanced configurations right? Could they be configured with hiera instead of saturating puppet-katello?

chris1984: We tried yesterday with the Hiera and it didnt work, @ekohl your thoughts happy to add it to Hiera but it didnt work during the demo when I tried.

chris1984:
Not needed anymore closing out

PR #239, opened 2018-03-15
Fixes #22907 - expose Pulp DB configuration params

mbacovsky:
This patch exposes most of the Pulp DB configuration and visible from installer.
The params are listed as advanced and not pollute installer --help.
I was able to install Katello with remote mongo with:
```
foreman-installer \
  --scenario katello \
  --katello-pulp-db-username=pulp_user \
  --katello-pulp-db-password=pulp \
  --katello-pulp-db-seeds="192.168.121.43:27017" \
  --katello-pulp-db-name="pulp_database" \
  --foreman-admin-password=changeme
```

### DIFF
--- a/manifests/pulp.pp
+++ b/manifests/pulp.pp
@@ -19,6 +19,19 @@ class katello::pulp (
   String $repo_export_dir_owner = $::katello::user,
   String $repo_export_dir_group = $::katello::group,
   Integer[0] $pulp_worker_timeout = $::katello::pulp_worker_timeout,
+  String $db_name = $::katello::pulp_db_name,
+  String $db_seeds = $::katello::pulp_db_seeds,
+  Optional[String] $db_username = $::katello::pulp_db_username,
+  Optional[String] $db_password = $::katello::pulp_db_password,
+  Optional[String] $db_replica_set = $::katello::pulp_db_replica_set,
+  Boolean $db_ssl = $::katello::pulp_db_ssl,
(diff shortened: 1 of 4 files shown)


manifests/pulp.pp
```diff
@@ -54,6 +66,19 @@
...
+    db_seeds               => $db_seeds,
+    db_username            => $db_username,
+    db_password            => $db_password,
+    db_replica_set         => $db_replica_set,
+    db_ssl                 => $db_ssl,
+    db_ssl_keyfile         => $db_ssl_keyfile,
+    db_ssl_certfile        => $db_ssl_certfile,
+    db_verify_ssl          => $db_verify_ssl,
+    db_ca_path             => $db_ca_path,
+    db_unsafe_autoretry    => $db_unsafe_autoretry,
```
ekohl: This parameter isn't in the class parameters and thus causing a failure.

mbacovsky:
@ekohl, thanks, I've fixed the problem and updated the tests slightly.
I was thinking the same about the pulp module, but I'm not sure about the consequences.
Exposing whole pulp would lead to renamed existing installer parameters and answers keys which would be better to do prior this is merged though.

iNecas:
Given we target this for 1.18, I would like to shoot for less with this PR, even at the price of needed to change the parameters later. Since I don't expect too many uses using this option, handling the change by migration + updating docs later should be enough after we actually expose puppet-pulp, which should be tracked as a separate issue.

PR #241, opened 2018-05-17
Fixes #23624 - make Apache directory options for /pub configurable

itsbill:
Hi,

These changes make the Apache directory options for /pub configurable via custom-hiera.yaml.
This is helpful when you need to disable directory listing for security scanner compliance.
By default nothing is changed, but if directory listing of /pub (both HTTP and HTTPS) were desired, adding the following line to `/etc/foreman-installer/custom-hiera.yaml` would prevent autoindex:
~~~
katello::pulp::pub_dir_options: "+FollowSymLinks"
~~~
Thank you in advance, and special thanks to @beav for all the help!

### DIFF
--- a/manifests/pulp.pp
+++ b/manifests/pulp.pp
@@ -32,6 +32,7 @@ class katello::pulp (
   Boolean $db_unsafe_autoretry = $::katello::pulp_db_unsafe_autoretry,
   Optional[Enum['majority', 'all']] $db_write_concern = $::katello::pulp_db_write_concern,
   Boolean $manage_db = $::katello::pulp_manage_db,
+  String $pub_dir_options = '+FollowSymLinks +Indexes',
 ) {
   include ::certs
   include ::certs::qpid_client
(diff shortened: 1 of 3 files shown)


chris1984:
This is the file we want to add them too, and you can see the examples here

https://github.com/theforeman/puppet-katello/blob/master/manifests/params.pp

ekohl:
> Looks good, my only change is that we should not hardcode values in the pulp.pp but we should have them as a param, and in the params.pp define the value.

They're already parameters. It's fine to hardcode a default - this allows users to override it in hiera as the author mentioned. Since this is an advanced feature I expect few people need, I'd slightly prefer this. Especially since we're going to change this up in https://github.com/theforeman/puppet-katello/pull/209 anyway. The `/pub` directory isn't really a pulp feature per se and if we're going to allow splitting the deployment, we should figure out where to host it.

That said, please do fix the quotes so our linter is happy.

itsbill:
Thank you all for the comments!  I force-pushed this version which references the issue and fixes the quotes.

PR #243, opened 2018-05-21
fixes #24070 - add registry to katello.yaml

thomasmckay:
Relies on https://github.com/Katello/katello/pull/7249

https://github.com/theforeman/puppet-katello_devel/pull/168
https://github.com/theforeman/puppet-katello/pull/243
https://github.com/Katello/katello/pull/7543

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -15,6 +15,7 @@ class katello::application (
   String $candlepin_oauth_key = $::katello::candlepin_oauth_key,
   String $candlepin_oauth_secret = $::katello::candlepin_oauth_secret,
   Stdlib::Httpsurl $pulp_url = $::katello::pulp_url,
+  Stdlib::Httpsurl $crane_url = $::katello::crane_url,
   String $qpid_url = $::katello::qpid_url,
   String $candlepin_event_queue = $::katello::candlepin_event_queue,
   Optional[String] $proxy_host = $::katello::proxy_url,
@@ -34,6 +35,7 @@ class katello::application (
   $post_sync_url = "${::foreman::foreman_url}${deployment_url}/api/v2/repositories/sync_complete?token=${post_sync_token}"
   $candlepin_ca_cert = $::certs::ca_cert
   $pulp_ca_cert = $::certs::katello_server_ca_cert
+  $crane_ca_cert = $::certs::katello_server_ca_cert
 
   foreman_config_entry { 'pulp_client_cert':
     value          => $::certs::pulp_client::client_cert,
--- a/templates/katello.yaml.erb
+++ b/templates/katello.yaml.erb
@@ -31,6 +31,11 @@
     :url: <%= @qpid_url %>
     :subscriptions_queue_address: <%= @candlepin_event_queue %>
 
+  # Internal configuration for communication from server to pulp crane service.
+  :container_image_registry:
+    :crane_url: <%= @crane_url %>
+    :crane_ca_cert_file: <%= @crane_ca_cert %>
+
 <%- unless [nil, :undefined, :undef, ''].include?(@proxy_host) -%>
   :cdn_proxy:
     :host: <%= @proxy_host %>
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -62,6 +62,7 @@ class katello::params {
   # Subsystems settings
   $candlepin_url = "https://${::fqdn}:8443/candlepin"
   $pulp_url      = "https://${::fqdn}/pulp/api/v2/"
+  $crane_url  = "https://${::fqdn}:5000"
 
   # database reinitialization flag
   $reset_data = 'NONE'
--- a/spec/classes/katello_application_spec.rb
+++ b/spec/classes/katello_application_spec.rb
@@ -21,6 +21,7 @@ describe 'katello::application' do
           :candlepin_oauth_key    => 'candlepin',
           :candlepin_oauth_secret => 'candlepin-secret',
           :pulp_url               => 'https://foo.example.com/pulp/api/v2/',
+          :crane_url              => 'https://foo.example.com:5000',
           :qpid_url               => 'amqp:ssl:localhost:5671',
           :candlepin_event_queue  => 'katello_event_queue',
           :proxy_host             => '',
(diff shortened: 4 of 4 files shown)


thomasmckay:
I don't like that the server errors out if there is something in the settings conf that is unrecognized, is that purposeful?

Also, will upgrades get the new stanza I've added or is there something else I need to do for that?

templates/katello.yaml.erb
```diff
@@ -31,6 +31,11 @@
     :url: <%= @qpid_url %>
     :subscriptions_queue_address: <%= @candlepin_event_queue %>
 
+  :registry:
+    :url: <%= @crane_url %>
+    :ca_cert_file: <%= @crane_ca_cert %>
+
+
```
ehelms: Nitpick - extra line here.

manifests/application.pp
```diff
@@ -15,6 +15,7 @@
   String $candlepin_oauth_key = $::katello::candlepin_oauth_key,
   String $candlepin_oauth_secret = $::katello::candlepin_oauth_secret,
   Stdlib::Httpsurl $pulp_url = $::katello::pulp_url,
+  Stdlib::Httpsurl $crane_url = $::katello::crane_url,
```
ehelms: The settings file generically referes to things as a regristry but this is specific to crane which is a bit of disconnect to me.

ehelms:
@thomasmckay SETTINGS is just a generic hash nothing special about it hence why it errors if something is missing. You'd need to write your code to handle the case where its not present or configured which is a good model to follow.

thomasmckay:
I thought about making it a crane section except I have plans to avoid crane completely at some point and also will add some registry-specific settings unrelated to crane. Perhaps I should change the settings themselves to reflect their related to crane.

Another question, should I just be using existing settings available in katello for crane comms? Maybe they are not needed at all.

@parthaa @jlsherrill - What do you think?

ehelms:
You could still make this generic to registry unless you think that will cause confusion.

ehelms:
BTW, this is what currently breaks nightly so arriving at a solution soon would be ideal.

ekohl:
I think the parameter should match the one in the config so `registry_url` would be best. I'm not a big fan of the generic `registry` though. `container_registry` would be better.

ekohl:
Ping?

thomasmckay:
@ekohl @ehelms - Updated PR and opened new issue to align with changes needed in katello

thomasmckay:
Updated. Pointers on tests?

stbenjam:
Please don't forget about puppet-katello_devel, which will need similar changes.

templates/katello.yaml.erb
```diff
@@ -31,6 +31,10 @@
     :url: <%= @qpid_url %>
     :subscriptions_queue_address: <%= @candlepin_event_queue %>
 
+  :container_image_registry:
+    :crane_url: <%= @crane_url %>
```
ekohl: Why are these prefixed with `crane_`? I'd expect them to be `url` and `ca_cert_file` (similar to others above).

templates/katello.yaml.erb
```diff
@@ -31,6 +31,10 @@
     :url: <%= @qpid_url %>
     :subscriptions_queue_address: <%= @candlepin_event_queue %>
 
+  :container_image_registry:
```
ekohl: I'm thinking about container image registry vs container registry but I'm not sure which would be better.

manifests/application.pp
```diff
@@ -34,6 +35,7 @@
   $post_sync_url = "${::foreman::foreman_url}${deployment_url}/api/v2/repositories/sync_complete?token=${post_sync_token}"
   $candlepin_ca_cert = $::certs::ca_cert
   $pulp_ca_cert = $::certs::katello_server_ca_cert
+  $crane_ca_cert = $::certs::katello_server_ca_cert
```
ekohl: This can probably also be a `Stdlib::Absolutepath`. I also think this should be kept together with the URL setting since they're related. We should have done the same with the other settings but that can be a separate PR.

templates/katello.yaml.erb
thomasmckay: These are the configs that katello uses to find the crane registry. These are not configs that are exposed to the users of the registry.

ekohl: Well, they are exposed through this module.

Also it doesn't answer why this is pinned to crane. If this option only supports crane and not a generic registry then I'd expect that the section is named `crane`. The alternative is that it's generic and any registry is supported. In both cases it doesn't make sense to me to prefix options with `crane_`.

thomasmckay: This section is to config the aspects of the newly implemented registry endpoint. As part of the implementation, katello needs to call pulp's crane; that's what these two configs are for. Removing the _crane_ prefix would then not convey what the url is meant to point at.

If preferred, I could add another entry to indicate the type of backend registry it is pointing at and then remove the _crane_ prefix from the existing two entries. Please suggest a name for this new entry.

templates/katello.yaml.erb
thomasmckay: I prefer as is but either is fine with me. We call the resource "container image" throughout the UI now (as we've tried to downplay the term _docker_).

templates/katello.yaml.erb
ekohl: It would help me to see how this section and its values are used. The PR mentioned in the opening message is merged but since then I think it has changed. Could you describe how it's used or link to the current code?

thomasmckay: Usage:
Don't expose registry if settings absent
https://github.com/Katello/katello/blob/master/app/controllers/katello/api/registry/registry_proxies_controller.rb#L410-L415

Pull objects from crane
https://github.com/Katello/katello/blob/master/app/lib/katello/resources/registry.rb#L22-L27

manifests/application.pp
thomasmckay: @ekohl I don't understand what is meant to change here.

ekohl: Ignore this one, I was mistakenly thinking that it was a class parameter.

thomasmckay:
I've edited the PR description to link all the relevant PRs. If there is more to do, please let me know. Thanks for your patience!

thomasmckay:
Is the following acceptable?
```
--- katello.yaml.example
 
  # Internal configuration for communication from server to pulp crane service.
  :container_image_registry:
    :backend_registry: crane  # must be crane
    :crane_url: https://localhost:5000
    :crane_ca_cert_file: /etc/pki/katello/certs/katello-server-ca.crt


--- puppet
    
  :container_image_registry:
    :backend_registry: crane  # must be crane
    :crane_url: <%= @crane_url %>
    :crane_ca_cert_file: <%= @crane_ca_cert %>
```

ekohl:
If `backend_registry` is not yet read by katello you can leave it out. It's just good to know where it's future proof and we don't need to change it again in a later release.

I'd also keep the comment in the puppet template. That way you have a minimal diff when you apply it for the first time.

thomasmckay:
I don't have merge permission in this repo, if someone else could do that.

chris1984:
Thanks @thomasmckay

PR #249, opened 2018-07-16
[WIP] Add Katello Apache tunables.

chris1984:
@ekohl started working on adding the 3 tuning options from the 6.2 tuning guide, i have not done tests yet. Let me know if it looks good and I will work on tests.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -74,6 +74,6 @@ class katello::application (
   }
 
   foreman::config::passenger::fragment{ 'katello':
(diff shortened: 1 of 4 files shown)


stbenjam:
In 6.2 can you set `apache::keepalive` in custom-hiera or will that not apply it globally?

ekohl:
@stbenjam without looking at the details: yes, it should apply globally. I'm not sure if we set it explicitly somewhere else but it should be easy to verify. Right now my goal is to get 1.19 RC prepared so I don't have that much time to do it myself.

PR #258, opened 2018-09-28
[WIP] Support syspurpose compliance event

jturel:


spec/classes/katello_qpid_spec.rb
```diff
@@ -44,7 +55,7 @@
             .with_ssl_key('/etc/pki/katello/private/foo.example.com-qpid-broker.key')
         end
 
-        ['entitlement.created', 'entitlement.deleted', 'pool.created', 'pool.deleted', 'compliance.created'].each do |binding|
+         @events.each do |binding|
```
stbenjam: You could maybe do `let` instead, `@events` doesn't seem to be defined here.

```ruby
let :events do
  [
    'entitlement.created',
    'entitlement.deleted',
    'pool.created',
    'pool.deleted',
    'compliance.created',
    'system_purpose_compliance.created'
  ]
end
```

jturel: That's what I tried the first time but let me put the let in a different spot and try again

spec/classes/katello_qpid_spec.rb
```diff
@@ -7,6 +7,17 @@
         facts
       end
 
+      let :events do
```
stbenjam: Ah I think I led you astray, let doesn't seem to work anywhere other than in an `it` block

From the test failure currently:

```
  `events` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).
```

jturel:
@stbenjam I cheated :)

PR #263, opened 2018-11-16
Drop keystore link in favor of standard Candlepin config

ehelms:


ehelms:
Goes along with https://github.com/theforeman/puppet-certs/pull/225

ehelms:
I can make that change from my testing. The change in certs is still needed as I want to move the keystore to the default Candlepin location and out of `/etc/pki/katello` intentionally to make each service and it's configuration more independent. Right now, you have to mount at least 4 different volumes to run Candlepin as a container. By centralizing these into `/etc/tomcat` and `/etc/candlepin` only two need to be mounted.

manifests/candlepin.pp
```diff
@@ -22,6 +22,7 @@
     oauth_secret                 => $oauth_secret,
     ca_key                       => $::certs::ca_key,
     ca_cert                      => $::certs::ca_cert_stripped,
+		keystore_file								 => $certs::candlepin::keystore_file,
```
ekohl: Could you fix tabs vs spaces here?

ehelms:
Updated and the change in puppet-certs has been merged

PR #265, opened 2018-11-19
Fixes #25511 - Disable max tasks by default to prevent memory issues.

chris1984:
https://projects.theforeman.org/issues/25511

### DIFF
--- a/spec/classes/katello_pulp_spec.rb
+++ b/spec/classes/katello_pulp_spec.rb
@@ -44,7 +44,7 @@ describe 'katello::pulp' do
             .with_enable_docker(true)
             .with_enable_ostree(false)
             .with_num_workers(1)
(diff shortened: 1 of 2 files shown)


chris1984:
[test prprocessor]

spec/classes/katello_pulp_spec.rb
```diff
@@ -44,7 +44,6 @@ class { '::katello':
             .with_enable_docker(true)
             .with_enable_ostree(false)
             .with_num_workers(1)
-            .with_max_tasks_per_child(2)
```
ekohl: You can still test this and verify the value is `nil` instead of `2`

chris1984:
@ekohl updated

evgeni:
I have no clue about the actual :sparkle: number 2

PR #266, opened 2018-11-21
Pulp vhosts80 moved to conf.d

ehelms:


ekohl:
Requires https://github.com/theforeman/puppet-pulp/pull/350

templates/pulp-apache.conf.erb
```diff
@@ -10,7 +10,7 @@ Alias /pub /var/www/html/pub
   Require all granted
 </Location>
 
-Include /etc/pulp/vhosts80/*.conf
+Include /etc/httpd/conf.d/vhosts80/*.conf
```
evgeni: `puppet-pulp` uses `${apache::confd_dir}`, but here we hardcode `/etc/httpd/conf.d`? (Debian ships httpd config in `/etc/apache2` ;))

evgeni:
Failures seem unrelated?

templates/pulp-apache.conf.erb
evgeni: this comment remains

ehelms:
Updated @evgeni

templates/pulp-apache.conf.erb
```diff
@@ -10,7 +10,7 @@ Alias /pub /var/www/html/pub
   Require all granted
 </Location>
 
-Include /etc/pulp/vhosts80/*.conf
+Include <%= scope['apache::confd_dir'] %>pulp-vhosts80/*.conf
```
evgeni: I'm not sure whether `apache::confd_dir` does or does not contain a trailing slash, so I'd prefer to be explicit here and add one before `pulp-vhost80`.

evgeni: https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/params.pp#L72 says it doesn't.

PR #267, opened 2018-11-26
Update qpid client certificate

ehelms:


ekohl:
Needs https://github.com/theforeman/puppet-certs/pull/229

ekohl:
Restarted the build now that the puppet-certs change has been merged. Perhaps we should also bump the major of puppet-certs already and set the minimum required version here.

ekohl:
https://github.com/theforeman/puppet-katello/pull/279 will be needed as well

PR #268, opened 2018-11-27
Use foreman::plugin

ehelms:


ehelms:
This change failed installation without https://github.com/theforeman/puppet-certs/pull/231 as it complained about not finding group 'foreman'. I am guessing this broke some unspoken dependency chain.

ehelms:
This is now no longer requires the puppet-certs change is good for review stand alone.

manifests/application.pp
```diff
@@ -49,30 +49,15 @@
...
-  package { $package_names:
-    ensure => installed,
-  } ->
-  file { "${::foreman::plugin_config_dir}/katello.yaml":
-    ensure  => file,
-    owner   => 'root',
-    group   => $::foreman::group,
-    mode    => '0640',
-    content => template('katello/katello.yaml.erb'),
-    notify  => [Class['foreman::service'], Foreman::Rake['db:seed'], Foreman::Rake['apipie:cache:index']],
```
ekohl: The seeding is a difference. With foreman we always do `database ~> plugin ~> service` where here we don't. That means it might get seeded with an invalid config if the package is already installed. Perhaps this is a bigger issue we need to address with katello's meta package anyway. For most plugins this isn't an issue because migrations typically don't use a config, but in Katello we do for Candlepin. That's why until now I've held off on this change.

ehelms: I know this passed in my testing. I can re-test this and report back.

ehelms: A fresh install and an upgrade test passed with this change. Perhaps we have just made the seeding more robust over time to not have the issue you were thinking of?

ehelms:
Rebased

ehelms:
@ekohl Any thoughts here?

ehelms:
Any further thoughts on why we shouldn't do this?

ekohl:
I'm still worried about the provisioning steps. Here we could end up with foreman db seed (with katello installed) -> katello configuration. Because Katello db seed scripts need candlepin to be configured, we could run into issues: https://github.com/theforeman/puppet-foreman/blob/022fcb91aa8072732bb29d494f990db3430ead56/manifests/init.pp#L369-L371

Perhaps this is something we should fix in puppet-foreman though.

ehelms:
I get your speculation of issues but what testing do you need to feel comfortable with this outside of an install and upgrade test?

ehelms:
@ekohl Should I close and abandon this?

ekohl:
I still like this effort and IMHO we should figure out how to make sure this works. Currently I don't have the cycles to investigate this too closely though. We could merge it to nightly if you have the time to deal with fallout.

ehelms:
I'm all about some fallout. When all available tests are passing what else is there to do?

ehelms:
@ekohl Do you know why we perform this -- https://github.com/theforeman/puppet-foreman/blob/master/manifests/init.pp#L372-L373 ?

ekohl:
No, see https://github.com/theforeman/puppet-katello/pull/268#discussion_r237118096. I'd be open to changing that part because logically the current implementation doesn't feel correct.

ehelms:
I'm gonna explore this further from a few different vectors.

ekohl:
Going to close this for now since it has merge conflicts. We're working on the other aspects and look back at this when those are finished.

PR #269, opened 2018-11-28
Drop apache user group config

ehelms:


ehelms:
Will require https://github.com/theforeman/puppet-foreman/pull/697 or else there are failures in the installer

ehelms:
I believe we can drop this as the original intent was to ensure that Pulp could read the CA key. But we no longer configure Pulp to need the CA key.

ekohl:
https://github.com/theforeman/puppet-katello/pull/270 should fix the tests

ehelms:
@ekohl Thanks for the test fixes. I re-ran this and it's now green.

ekohl:
Another potential reason was the legacy export. IIRC Katello does some operations on the Pulp files directly, for example in the old export. While this is a bad design that I'd love to get rid of, I'm not sure we should be doing this change before 3.10 is branched.

ehelms:
The original message when this was added was:

```
    adding apache user to the foreman group

    so pulp can read the candlepin private key
``

To me this has naught to do with export. Merging this is only a concern if we plan to release all of the puppet-katello changes into 3.10 which, as I understand things, we do not. A full nightly pipeline did pass with this change. I ran a test of export with these changes and all was good.

PR #272, opened 2018-12-12
Simplify package name handling

ekohl:
f614db22fdb78f67b2e778f1729512683de4a251 removed the use of $rubygem_katello_ostree variable, but not its definition.

We also don't want to hard fail because we sometimes want to compile on other distros. The most common use case is CI testing our installer by running it with --help. Since that happens on Travis (which is Ubuntu) it hard fails.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,25 +1,12 @@
 # Katello Default Params
 class katello::params {
 
-  case $::osfamily {
-    'RedHat': {
-      case $::operatingsystem {
-        'Fedora': {
-          $rubygem_katello = 'rubygem-katello'
-          $rubygem_katello_ostree ='rubygem-katello_ostree'
-        }
-        default: {
-          $rubygem_katello = 'tfm-rubygem-katello'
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -1,24 +1,10 @@
...
-          $rubygem_katello = 'rubygem-katello'
-          $rubygem_katello_ostree ='rubygem-katello_ostree'
-        }
-        default: {
-          $rubygem_katello = 'tfm-rubygem-katello'
-          $rubygem_katello_ostree ='tfm-rubygem-katello_ostree'
-        }
-      }
-
-      $package_names = ['katello', $rubygem_katello]
```
ehelms: You forgot this

ehelms:
Looks like some new test failures:

```
Failure/Error: it { expect { should contain_class('katello') }.to raise_error(Puppet::Error, /#{facts[:hostname]}: This module does not support osfamily #{facts[:osfamily]}/) }
```

PR #276, opened 2019-01-09
Redirect Docker pull to Crane

ehelms:


templates/pulp-apache-ssl.conf.erb
```diff
@@ -11,3 +11,7 @@ Alias /pub /var/www/html/pub
   </IfModule>
   Options <%= @pub_dir_options %>
 </Location>
+
+# Crane redirects
+Redirect /v1 https://<%= scope['fqdn'] %>:5000/v1
```
evgeni: Can we avoid using the `fqdn` here? Otherwise this will break (unsupported) multi-name/multi-home setups.

evgeni: If I read the [docs](https://httpd.apache.org/docs/trunk/rewrite/remapping.html) correctly, the following should work:
```
Redirect /v1 https://%{HTTP_HOST}:5000/v1
Redirect /v2 https://%{HTTP_HOST}:5000/v2
```

ehelms: What are you suggesting we use instead?

evgeni: @ehelms my second comment :)

mirzal: @evgeni After adding this to relevant httpd config file, httpd will start sending header: `Location: https://%{HTTP_HOST}:5000/v1`. Looks like Redirect comes from [mod_alias](https://httpd.apache.org/docs/trunk/mod/mod_alias.html), which doesn't understand [mod_rewrite](https://httpd.apache.org/docs/trunk/mod/mod_rewrite.html#rewritecond) variables.

Following passed simple check:
```
RewriteEngine on
RewriteRule ^/v1/(.*)$ https://%{HTTP_HOST}:5000/v1/$1 [R=301,L]
RewriteRule ^/v2/(.*)$ https://%{HTTP_HOST}:5000/v2/$1 [R=301,L]
```

But:
1. Unconditionally enabling mod_rewrite in the middle of complex config doesn't sound like a good idea to me.
2. It won't catch requests to $HOST/v1 (without trailing slash). Not sure if they are meaningful to begin with.

Quick search over web shows that if we don't want to use mod_rewrite, we might use another virtual host or proxy directive. I have zero experience with these, so I haven't tried.

ehelms: @evgeni Any thoughts to @mirzal comment?

evgeni: First thought: "well, fuck"
Second thought: "but [mod_alias docs](https://httpd.apache.org/docs/trunk/mod/mod_alias.html) say I can use any [expression](https://httpd.apache.org/docs/trunk/expr.html) syntax I want"
Third thought: "well, I am off for some testing"

evgeni: So this expression syntax works starting with 2.4.19, but we only have 2.4.6 in EL7 :/
I'll go hunt for more ideas later.

evgeni: Okay, so my next idea was to push these into a `<Location>` directive, but [the docs say that this is not supported](https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html#rewriterule):
Although rewrite rules are syntactically permitted in *Location* and *Files* sections (including their regular expression counterparts), this should never be necessary and is unsupported. A likely feature to break in these contexts is relative substitutions.

Now looking at the bigger picture, the snippet in question is deployed into `/etc/httpd/conf.d/05-foreman-ssl.d/pulp.conf` on the Katello box, which gets included in `/etc/httpd/conf.d/05-foreman-ssl.conf`

The main config is using `mod_rewrite` in another `<Directory>`, so I kinda think it should be safe to also enable it here. But you never know until you really try.

ekohl: Do we want to redirect or would we want to reverse proxy?

ehelms:
@mirzal @evgeni Tested and worked for me with rewrite update, please review

jlsherrill:
I'm not sure this is the correct approach, or that we should be redirecting to crane at all.  Let me do some investigation and i will get back with you

ehelms:
Good eye @jlsherrill  - turns out this should have worked all along via the Ruby application code but a rogue header was breaking things! I am therefore closing these PRs in favor of a solution inside Ruby.

PR #282, opened 2019-03-12
Refs #26323 - Remove max_tasks_per_child setting

chris1984:


ekohl:
This needs a migration on the installer side as well to remove the answer.

chris1984:
Migration is here https://github.com/theforeman/foreman-installer/pull/327

PR #294, opened 2019-06-26
Support new global proxy for content

jlsherrill:


jlsherrill:
as part of pulp3 support we are moving to utilizing the HttpProxy objects for syncing, subscription actions, etc...   For at least a couple releases, i wanted to maintain the existing installation options for setting the global proxy.  So for now this will call a rake task which creates/updates the global content proxy and sets the setting which selects it.

We will need to maintain the setting of the pulp2 configuration until we complete moving over to this new mechanism (should just be a couple more weeks)

jlsherrill:
also, this is very rough and untested, opening now to get feedback

jlsherrill:
@ekohl would you have any time to look at this soon?

jlsherrill:
@ekohl so i saw this as a temporary measure to give consistency over a short term (1 to 2 releases).  It will also let us work on migrating the various pieces over to the new setting in chunks without breaking production.  The idea is that at least for the next 1 to 2 releases, the installer is still the way you set the proxy (which is how our documentation spells it out today).

Another idea i had was to not set it in puppet but instead do it via a post hook, such that it only runs when the user explicitly passes in the params (that way its not executed on every run).

Either way, we'd probably yank the installer method for setting the proxy fairly quickly.

jlsherrill:
we could populate this based on existing config data (https://github.com/theforeman/puppet-katello/blob/master/templates/katello.yaml.erb#L38) in seeds i guess, that does seem better.  Thanks for the brainstorming!

PR #297, opened 2019-08-12
No longer configure pulp2 global proxy

jlsherrill:
since we're now configuring the repositories directly


jlsherrill:
It is, thats currently how we pass the installer values for --katello-proxy to the application.  

The idea is that eventually we remove those options from the installer completely and just configure it within the application, but the thought was to let all that bake for a couple releases until we remove the installer options (as lots of docs/scripting depend on the installer)

PR #299, opened 2019-09-12
Fixes #27845 - quote values in katello.yaml

ekohl:
Some values may contain special characters that have a meaning in YAML.  By quoting this is avoided.


ehelms:
Tests need updating to include these new quotes

ekohl:
Updated the tests.

PR #304, opened 2019-11-29
Allow management of single products and add manage_httpd parameter for pulp

laugmanuel:
This PR adds the feature to only manage some parts of the installation (Katello, Pulp, Candlepin, Qpid) which is needed for a distributed setup.
It also introduces the `manage_httpd` parameter to `::katello::pulp` to enable the management of Apache for a standalone Pulp installation.

The `http_manage` parameter is not exposed in the `init.pp` for now. For that reason it is necessary to call the module as follows if one want's to use the parameter:
```
class { '::katello':
  manage_candlepin   => false,
  manage_qpid        => false,
  manage_pulp        => false, # explicitly declared below to overwrite parameters
  manage_application => false,
  qpid_hostname      => 'qpid.example.com',
  proxy_url          => 'http://proxy.example.com',
  proxy_port         => 8000,
}

class { '::katello::pulp':
  manage_httpd => true,
}
```

The originating PR is https://github.com/theforeman/puppet-katello/pull/218

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -192,11 +192,14 @@ class katello (
 ) inherits katello::params {
 
   include katello::repo
-  if $manage_candlepin {
-    include katello::candlepin
-  }
   if $manage_qpid {
     include katello::qpid
+  }
+  if $manage_candlepin {
+    include katello::candlepin
+    if $manage_qpid {
+      Class['katello::qpid'] -> Class['katello::candlepin']
+    }
   }
   if $manage_pulp {
     include katello::pulp
@@ -205,8 +208,8 @@ class katello (
   if $manage_application {
     include katello::application
     Class['katello::repo'] -> Class['katello::application']
-    if $manage_qpid and $manage_candlepin {
-      Class['katello::qpid'] -> Class['katello::candlepin'] -> Class['katello::application']
+    if $manage_candlepin {
+      Class['katello::candlepin'] -> Class['katello::application']
     }
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -52,6 +52,14 @@
 #
 # $package_names::      Packages that this module ensures are present instead of the default
 #
+# $manage_application:: Whether to manage the katello application part (for distributed setups).
+#
+# $manage_qpid::        Whether to manage the QPID part (for distributed setups).
+#
+# $manage_pulp::        Whether to manage the Pulp part (for distributed setups).
+#
+# $manage_candlepin::   Whether to manage the Candlepin part (for distributed setups).
+#
 # $manage_repo::        Whether to manage the yum repository
 #
 # $repo_version::       Which yum repository to install. For example
@@ -147,6 +155,10 @@ class katello (
   Boolean $enable_docker = $katello::params::enable_docker,
   Boolean $enable_deb = $katello::params::enable_deb,
 
+  Boolean $manage_application = $katello::params::manage_application,
+  Boolean $manage_qpid = $katello::params::manage_qpid,
+  Boolean $manage_pulp = $katello::params::manage_pulp,
+  Boolean $manage_candlepin = $katello::params::manage_candlepin,
 
   Stdlib::Absolutepath $repo_export_dir = $katello::params::repo_export_dir,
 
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -180,12 +192,21 @@
...
+  }
+  if $manage_pulp {
+    include katello::pulp
+    Class['katello::repo'] -> Class['katello::pulp']
+  }
+  if $manage_application {
+    include katello::application
+    Class['katello::repo'] -> Class['katello::application']
+    if $manage_qpid and $manage_candlepin {
+      Class['katello::qpid'] -> Class['katello::candlepin'] -> Class['katello::application']
```
ekohl: I think this should be split: `Class['katello::qpid'] -> Class['katello::candlepin']` should be under `$manage_candlepin` as a separate `if $manage_qpid`. Then this if block should just be `if $manage_candlepin`

laugmanuel: Thanks - I simplified the ordering of classes in the following commit.

manifests/init.pp
```diff
@@ -180,12 +192,22 @@
...
-  include katello::qpid
-  include katello::pulp
-  Class['katello::repo'] -> Class['katello::pulp']
-  include katello::application
-  Class['katello::repo'] -> Class['katello::application']
-  Class['katello::qpid'] -> Class['katello::candlepin'] -> Class['katello::application']
-
+  if $manage_candlepin {
+    include katello::candlepin
+    Class['katello::qpid'] -> Class['katello::candlepin']
```
ekohl: Does this work if the actual inclusion is later?

laugmanuel:
I do support your thoughts on separating the classes wherever possible and make them less or not dependent on each other. This would favour distributed installations a lot.
However we should try not to break the current behaviour of this module. For this purpose the changes above (manage_*) would be good because we could _hide_ the calls to all the other (standalone) classes.

manifests/init.pp
laugmanuel: I'm just wondering why we need the relationship definitions in the first place?
By default Puppet applies resources in the order they are declared in the manifest. Then there would be no need to account for every combination - we just need to make sure the ordering of includes is correct in the first place.
Is there something special here that I'm missing?

ekohl: > By default Puppet applies resources in the order they are declared in the manifest.

This is in no way a guarantee. However, Candlepin will fail to start if qpid is not available and configured. More specifically: it needs the event queue to be defined which we do in `katello::qpid`.

Part of this is old school Puppet. Prior to Puppet 4 the order was actually randomized and you had to be explicit. However, this particular case is actually explicitly a requirement and the app is known to fail hard if the condition is not met.

laugmanuel: Even though it should be no problem to declare relationships between resources before declaring the resources itself, I did change the ordering of candlepin and qpid to make the order consistent with the relations.

> Relationships are not limited by evaluation-order; you can declare a relationship with a resource before that resource has been declared.
(from https://puppet.com/docs/puppet/6.10/lang_relationships.html)

ekohl:
I've been thinking about this for a while and was actually very close to implementing it.

https://github.com/theforeman/puppet-katello/pull/306 is an implementation that automatically sets up the dependencies allowing to just include what you need and it'll automatically compose all components in the right order.

The next step would be to make the classes inherit from `katello::params` so they can be individually included. In `init.pp` it would call every class explicitly with all parameters to ensure the same values are used. In a distributed setup you would simply not use the main `katello` class but build your own composition.

What are your thoughts on this?

ekohl:
I've now implemented https://github.com/theforeman/puppet-katello/pull/308. For now it's only spec tested and there are some things I'm moving around that I don't know for sure about. I would very much like comments on the design while we merge the commits leading up to it (which are safe and can already be merged).

Next I'm going to look at using these in the installer whether that also makes sense. Another is to merge in https://github.com/theforeman/puppet-foreman_proxy_content so this module contains all the profiles/roles you want for the Katello ecosystem.

laugmanuel:
@ekohl, thank's for the initial implementation of the automatic component composing. I will have a look at #306 , #307 and #308 later and report back in the respective PR.

What are your plans for this PR? Do you mind merging it as an intermediate solution for #308 ? As far as I'm concerned I would very much appreciate that.

ekohl:
> What are your plans for this PR? Do you mind merging it as an intermediate solution for #308 ? As far as I'm concerned I would very much appreciate that.

No, I don't intend on merging an intermediate solution. A large issue in our installer is that we already have a lot of parameters and this only makes it harder. It can also easily break things. 

There's a few things I want to figure out:

https://github.com/theforeman/puppet-katello/blob/44741e4f078a86987774b84b6e366a5b732b7091/manifests/qpid.pp#L26

This looks odd to me. I strongly suspect this should actually be in `katello::qpid_client` instead. When you start splitting up the deployment, that can matter.

https://github.com/theforeman/puppet-katello/blob/44741e4f078a86987774b84b6e366a5b732b7091/manifests/pulp.pp#L88-L102

I think this is needed by Katello rather than Pulp. The permissions on `/var/lib/pulp` also can be an issue in the Pulp 2 to 3 migration.

The benefit of #308 is that we expose the parts in such a way that they're easy to use as separate components but still enforce a monolithic install in our installer.

I realize the Pulp part still needs a parameter to change the vhost handling which is a hard problem to solve. Currently looking into that.

laugmanuel:
Closed in favor of https://github.com/theforeman/puppet-katello/pull/312

PR #305, opened 2019-11-29
Use the nss cert name from certs::qpid

ekohl:
Rather than hardcoding it, the certs class that actually creates these should be responsible for naming it.

Draft because it depends on https://github.com/theforeman/puppet-certs/pull/265


evgeni:
should this bump the requirement on `katello/certs`?

ekohl:
It should. I'll look at an actual release.

ekohl:
Updated, depends on https://github.com/theforeman/puppet-certs/pull/271

PR #306, opened 2019-11-29
Implement automatic dependency chaining

ekohl:
While it's not possible to collect classes, it is possible to collect anchors. This allows including it anywhere and the right dependency will be set.

This is used to simplify the main class. As an example, the katello::repo is extracted. Ideally all classes become standalone so they can be exposed to the user in the installer. It also makes it easy to do split deployments.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -52,7 +52,7 @@ class katello::application (
   Class['certs', 'certs::ca', 'certs::qpid'] ~> Class['foreman::plugin::tasks']
 
   # Katello database seeding needs candlepin
-  Anchor <| title == 'katello::repo' |> ->
+  Anchor <| title == 'katello::repo' or title ==  'katello::candlepin' |> ->
   package { $package_names:
     ensure => installed,
   } ->
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -52,6 +52,7 @@ class katello::application (
   Class['certs', 'certs::ca', 'certs::qpid'] ~> Class['foreman::plugin::tasks']
 
   # Katello database seeding needs candlepin
+  Anchor <| title == 'katello::repo' |> ->
   package { $package_names:
     ensure => installed,
   } ->
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -46,7 +46,8 @@ class katello::candlepin (
     db_ssl_verify                => $db_ssl_verify,
     manage_db                    => $manage_db,
     subscribe                    => Class['certs', 'certs::candlepin'],
-  }
+  } ->
+  anchor { 'katello::candlepin': }
 
   contain candlepin
 }
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -16,6 +16,7 @@ class katello::candlepin (
   include certs
   include certs::candlepin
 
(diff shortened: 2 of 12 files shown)


manifests/application.pp
```diff
@@ -60,6 +60,7 @@
   Class['certs', 'certs::ca', 'certs::qpid'] ~> Class['foreman::plugin::tasks']
 
   # Katello database seeding needs candlepin
+  Anchor <| title == 'katello::repo' or title ==  'katello::candlepin' |> ->
```
ehelms: Can you expand on the choice of `anchor` over `contain` ? I'm trying to better understand what this code is doing and my first read of things leads to Anchor being compatibility code for older Puppets?

ekohl: `contain` equals `include` but also keep the resources in this class. Let's take the following code:
```puppet
class qpid {
  package { 'qpid': }
}

class candlepin {
  include qpid
  package { 'candlepin': }
}

class katello {
  require candlepin # equivalent to include candlepin + Class['candlepin'] -> Class['katello']
  package { 'katello': }
}
```
Now in the catalog the qpid resource can float. That means if you include katello you can be sure that candlepin is installed before katello but qpid might be installed after katello. That means it's valid to say `Class['katello'] -> Class['qpid']`.

However, if you used `contain qpid` instead then the `Class['qpid']` is considered part of `Class['candlepin']`. You know all resources within that class are executed before `Class['katello']`. Now `Class['katello'] -> Class['qpid']` will fail to compile because it a cyclic dependency.

Another benefit can be that it propagates refreshes. Suppose you have the following:
```puppet
class qpid {
  contain qpid::install
  contain qpid::config
  contain qpid::service
}

Class['qpid'] ~> Class['katello']
```

This means that if the `qpid` service is restarted, then so is `katello`. This can be a benefit or a downside. Using `contain` allows using this.

However, here we want to achieve something that **if** a resource is present, it should be before x. If it's not, then it's fine. That allows using:
```puppet
include katello
include katello::repo
```
Puppet will chain the resources correctly regardless.

We're using `anchor` here. In the past that was using to achieve `contain` behavior. `anchor` is a concrete resource and was always part of a class. However, that was basically an abuse of anchor to achieve a goal.

Anchors can be used to define an API. In this case we create a marker to signal qpid is provisioned while still allowing it to be optional. In our installer case that's currently not relevant, but it'll make it easier in the future. https://github.com/theforeman/puppet-katello/pull/304 shows there is demand for split installations already.

We use anchors here because you can't [collect](https://puppet.com/docs/puppet/latest/lang_collectors.html) a class. We could also collect specific resources (`exec`, `file`) but `anchor` is an abstract thing. For example, we can use it also to tag a certificate generation is done.

This is a cleaner implementation than what we have in Foreman (https://github.com/theforeman/puppet-puppet/blob/964f110cdb8e654572b96553607e0784674d2a8e/manifests/server/install.pp#L5-L12 and https://github.com/theforeman/puppet-foreman/blob/36aa5389b7da41d4770fa61a1c10a78239695735/manifests/config.pp#L3-L7 and https://github.com/theforeman/puppet-foreman_proxy/blob/cd99718f813c12acfb0b14676780fa8883ce1092/manifests/config.pp#L4-L15).

ekohl:
FYI: this is marked as backwards incompatible because it's dropping the repo parameters from init.pp. Otherwise it's fully compatible.

PR #310, opened 2019-12-10
Drop creating /var/lib/pulp

ehelms:


ehelms:
This being here can cause problems such as -- https://github.com/theforeman/puppet-pulpcore/issues/17

manifests/pulp.pp
```diff
@@ -86,19 +86,11 @@
...
   file { $repo_export_dir:
-    ensure => directory,
-    owner  => $repo_export_dir_owner,
-    group  => $repo_export_dir_group,
-    mode   => '0755',
+    ensure  => directory,
+    owner   => $repo_export_dir_owner,
+    group   => $repo_export_dir_group,
+    mode    => '0755',
+    require => Package['pulp-server'],
```
ekohl: Might be cleaner to require `Class['pulp']`
```suggestion
    require => Class['pulp'],
```

PR #311, opened 2019-12-20
Use Stdlib::Httpurl datatype to support setting a proxy protocol

laugmanuel:
We need to support Strings as well because we might need the protocol for the proxy.
E.g. this Katello seed will fail if the proxy is specified without a protocol: https://github.com/Katello/katello/blob/master/db/seeds.d/115-http_proxy.rb#L3

The datatype was changed in https://github.com/theforeman/puppet-katello/pull/308

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -21,7 +21,7 @@
 class katello::application (
   Integer[0] $rest_client_timeout = 3600,
   Optional[Enum['SSLv23', 'TLSv1', '']] $cdn_ssl_version = undef,
-  Optional[Stdlib::Host] $proxy_host = undef,
+  Optional[Stdlib::HTTPUrl] $proxy_host = undef,
   Optional[Stdlib::Port] $proxy_port = undef,
   Optional[String] $proxy_username = undef,
   Optional[String] $proxy_password = undef,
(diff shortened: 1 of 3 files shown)


manifests/application.pp
```diff
@@ -21,7 +21,7 @@
 class katello::application (
   Integer[0] $rest_client_timeout = 3600,
   Optional[Enum['SSLv23', 'TLSv1', '']] $cdn_ssl_version = undef,
-  Optional[Stdlib::Host] $proxy_host = undef,
+  Optional[Variant[Stdlib::Host, String]] $proxy_host = undef,
```
ekohl: So previously this was a `Stdlib::Httpurl` and it may have been my mistake that I didn't change it. I don't know if Katello actually accepts a plain host. Probably easier to revert back to that.
```suggestion
  Optional[Stdlib::Httpurl] $proxy_host = undef,
```

ekohl:
Looking a bit further, it appears they're used in seeds (https://github.com/Katello/katello/commit/9a926671f29f351939d34005a65cda8b9d911f00) where it's called as `URI(config[:host])`. That simply doesn't work for just a host.

laugmanuel:
Sorry I didn't catch this one. Fix it in `init.pp` as well and rebased the PR.

laugmanuel:
Should be ready now 👍

ekohl:
I was *just* looking at it :)

PR #313, opened 2020-01-16
Refs #28695: Add Pulp 2 migration settings

ehelms:


manifests/init.pp
```diff
@@ -156,12 +159,14 @@
...
-    enable_puppet => $enable_puppet,
-    enable_docker => $enable_docker,
-    enable_deb    => $enable_deb,
+    enable_ostree         => $enable_ostree,
+    enable_yum            => $enable_yum,
+    enable_file           => $enable_file,
+    enable_puppet         => $enable_puppet,
+    enable_docker         => $enable_docker,
+    enable_deb            => $enable_deb,
+    use_pulp_2_for_file   => $use_pulp_2_for_file,
```
ekohl: These probably need to be parameters to `katello::application`, not globals.

ehelms: I honestly have no idea what's what anymore. So if thats the best place for this I'll do whatever. `katello::application` includes params which inherits globals. So when and how do you know where to pass what?

ehelms:
I saw these errors locally and I struggled to decipher the issue given its similar to other parameters.

manifests/init.pp
```diff
@@ -125,6 +129,9 @@
   Boolean $enable_docker = true,
   Boolean $enable_deb = true,
 
+  Boolean $use_pulp_2_for_file = $use_pulp_2_for_file,
```
ekohl: This needs to be an explicit value, such as:
```suggestion
  Boolean $use_pulp_2_for_file = true,
```

manifests/init.pp
ekohl: I should write this down in the README, but the general rule is that `katello::globals` affects multiple services that a user is expected to be able to change (https://github.com/theforeman/foreman-installer/pull/421 would expose it to the end user). `katello::params` is intended for advanced uses and the installer wouldn't expose them. Then we have the individual classes (`katello::application`, `katello::candlepin`, `katello::pulp` and `katello::qpid`) which are directly exposed to the user as top level parameters. The class `katello` is for compatibility and going away. Right now it's there not to break the installer experience. The reason it's still here is more for downstream than upstream.

manifests/init.pp
ehelms: :man_facepalming: Duh Eric

ehelms:
@jlsherrill if this is merged, but Pulp 3 is not setup, will Katello bomb out? Or is this safe to merge ahead of itme?

jlsherrill:
This should be safe to merge ahead of time

PR #315, opened 2020-01-22
Fixes #28837 - removes cdn_proxy from config

jjeffers:
This change removes the cdn proxy and HTTP proxy configurations.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -6,18 +6,6 @@
 # @param cdn_ssl_version
 #  SSL version used to communicate with the CDN
 #
-# @param proxy_host
-#  URL of the proxy server
-#
-# @param proxy_port
-#  Port the proxy is running on
-#
-# @param proxy_username
-#  Proxy username for authentication
-#
-# @param proxy_password
-#  Proxy password for authentication
-#
 # @param use_pulp_2_for_file
 #  Configure Katello to use Pulp 2 for file content
 #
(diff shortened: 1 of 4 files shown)


jlsherrill:
lets hold off on merging this until we have an upgrade decision

jjeffers:
> Please also remove them from `katello.pp`

@ekohl apologies but not familiar with this file. Can you point me in the right direction?

ekohl:
That's probably because I pointed you in the wrong direction. I meant https://github.com/theforeman/puppet-katello/blob/master/manifests/application.pp.

jlsherrill:
alright, i think we have a good path forward, lets proceed with this when its green

ekohl:
The tests will need updating.

It would be nice if there was a bit of context. What will be the recommended alternative?

jjeffers:
> The tests will need updating.
> 
> It would be nice if there was a bit of context. What will be the recommended alternative?

Looks like a previous cron build failed for similar problems https://travis-ci.org/theforeman/puppet-katello/jobs/640209707?utm_medium=notification&utm_source=github_status

I'm confused wrt the earlier change in master passed, but that one failed.

ekohl:
Ah yes, you're right. https://github.com/theforeman/puppet-katello/pull/316

jlsherrill:
@ekohl so previously we were trying to:
1) maintain compatibility in the installer
2) support 'upgrades', so that any proxy configured would flow into the new application hosted data/setting that is now there for a 'global content proxy server'

The solution was to use this katello.yaml config, plus a seeds file, to set the config at seeds time.  The dots that weren't connected were that upon upgrade, updating the setting resulted in katello attempting to change the proxy configuration for repositories configured in pulp.  This did not work since pulp was not available at seeding time (apache wasn't started).  So after some discussions, we decided to remove these options and not worry about 1).

Now how will we fix 2)?   The idea is that we will leave these settings in the answers file and write an upgrade task in katello that will read the answers file and configure the global proxy in katello.  Since all this has been up upstream for some time (which i'm not sure why it hasn't caused an issue), i'm imagining this would only be a downstream only change.  apologies for mentioning downstream  ;)   I'm also open to other solutions

PR #319, opened 2020-01-29
Fixes #28888 - add cert auth for docker registry

jlsherrill:


jlsherrill:
Yep, this /v2/ api is the docker registry that katello provides (that is backed by pulp).  These end points need to support cert-based auth in order for a foreman proxy content to be able to sync from it.

jlsherrill:
Thanks @wbclark !

PR #321, opened 2020-02-04
Refs #28922: Communicate via localhost to Candlepin

ehelms:


ehelms:
I added a `candlepin_host` parameter and used that as the driver for configuration as it seemed more natural than going through a certs option

manifests/application.pp
```diff
@@ -55,7 +55,7 @@
   $enable_deb = $katello::params::enable_deb
   $pulp_url = $katello::params::pulp_url
   $pulp_ca_cert = $certs::katello_server_ca_cert # TODO: certs::apache::...
-  $candlepin_url = $katello::params::candlepin_url
+  $candlepin_url = "https://${katello::candlepin_host}:8443/candlepin"
```
ekohl: The `katello::application` should not assume the `katello` class exists. I should write a PR, but after 2.1 I want to deprecate it altogether. Anything that's shared between `katello::application` and `katello::candlepin` should go through `katello::params` (as the URL was previously). You can then remove the parameter from `init.pp`

ehelms: I think I follow what you are suggesting. So check out this line and change now... and let me know if I got there or not.

ekohl:
This is a valid way, but if you look at https://github.com/theforeman/foreman-installer/pull/421 there's a few layers:
* `katello::global` - all parameters on this class are exposed to installer users as `--global-*`
* `katello::params` - these are shared parameters between multiple components but not exposed to installer users. They are intended for advanced users (directly using the Puppet modules or via custom-hiera.yaml).
* application layer - all parameters are exposed to installer users
  * `katello::candlepin` maps to `--candlepin-*`
  * `katello::pulp` maps to `--pulp-*`
  * `katello::application` maps to `--application-*`

Now for the installer we don't want to make this configurable while still allowing advanced users to change it. That means it's a parameter on `katello::params`. It would be confusing to have the parameters to that again at the application layer so on `katello::candlepin` there would not be a `$host` parameter but rather inside the body it'd have `$body = $katello::params::candlepin_host`. And the same for `katello::application`.

This way we avoid having installer parameters that the user is not supposed to touch while still providing all the power to advanced users who want to do custom things.

Right now I feel this design is still rather experimental and we're inventing a new pattern. This means we may find out it doesn't work or doesn't scale, but I'd like to try it because I feel it has the potential to strike the balance well.

jturel:
I tested the trio of PRs and it works nicely. Thanks!

ehelms:
OK... I think I get the gist of the idea now. Let's see what you think. The tests here won't pass until I merge the corresponding Candlepin and Certs changes.

manifests/params.pp
```diff
@@ -21,11 +23,12 @@
...
 #   The oauth secret for Candlepin
 class katello::params (
-  Stdlib::Httpsurl $candlepin_url = "https://${facts['fqdn']}:8443/candlepin",
   Stdlib::Httpsurl $pulp_url = "https://${facts['fqdn']}/pulp/api/v2/",
   Stdlib::Httpsurl $crane_url = "https://${facts['fqdn']}:5000",
   Stdlib::Host $qpid_hostname = 'localhost',
   String[1] $candlepin_oauth_key = $katello::globals::candlepin_oauth_key,
   String[1] $candlepin_oauth_secret = $katello::globals::candlepin_oauth_secret,
+  Stdlib::Host $candlepin_host = 'localhost',
+  Stdlib::Httpsurl $candlepin_url = "https://${candlepin_host}:8443/candlepin"
```
ekohl: I don't know if this works. I've seen people use it, but never trusted it. Given this is the only way to deploy candlepin, I'd move the `$candlepin_url` part to the body of this class.

.fixtures.yml
```diff
@@ -24,6 +24,7 @@ fixtures:
       puppet_version: '>= 6.0.0'
     stdlib:        "https://github.com/puppetlabs/puppetlabs-stdlib.git"
     systemd:       "https://github.com/camptocamp/puppet-systemd.git"
+    transition:    "https://github.com/puppetlabs/puppetlabs-transition.git"
```
ekohl: Nit: I'd do this as a separate commit. Fine to do it in the same PR.

manifests/params.pp
ehelms: This worked in my testing running these changes live with the installer. If its defined in the body, will other modules be able to access it?

.fixtures.yml
ehelms: Split out

manifests/params.pp
ekohl: Yes, that still works.

ehelms:
If I recap what you are saying, this change is a breaking change for
puppet-candlepin? I'm not sure I follow why that is given we are adding a
feature and changing a default. What is the breaking change?

ekohl:
> This also means increasing the minimum required version of the candlepin module. Perhaps the candlepin PR should bump the major version and this should update metadata.json.

The exact candlepin version doesn't matter that much, but the lower bound should set a version that's actually compatible:
https://github.com/theforeman/puppet-katello/blob/f6dda3ec51cadb214206cd00b042db16c1d4b76f/metadata.json#L25

Right now puppet-candlepin's master branch has version 8.0.0 which equals the already released version. In the past we just merged these things and worked it out around branching but I'd like to get in the habit of fixing it while we're making changes so I'm not stuck with a bunch of work around branching. https://github.com/theforeman/puppet-candlepin/pull/146 & https://github.com/theforeman/puppet-katello/pull/325 work towards that.

PR #322, opened 2020-02-07
Fixes #28961: Drop crane configuration from katello.yaml if using Pulp 3

ehelms:


ehelms:
Related https://github.com/theforeman/puppet-foreman_proxy_content/pull/239

ehelms:
@jlsherrill Thoughts on this cleanup? We could also wait till 2.1 Foreman release to drop Crane, the external proxy situation complicates things since we are planning for it to be a Pulp 2 only deployment for a couple releases.

templates/katello.yaml.erb
```diff
@@ -33,7 +33,9 @@
     :url: <%= @qpid_url %>
     :subscriptions_queue_address: <%= @candlepin_event_queue %>
 
+  <%- if @use_pulp_2_for_docker -%>
```
ekohl: Should this be a line higher so you don't have an empty line at the end of the file?

ehelms:
Closing in favor of https://github.com/theforeman/puppet-katello/pull/356

PR #324, opened 2020-02-17
Refs #29039 - Use foreman's Apache config standalone

ekohl:
By using only the Apache config, it's possible to use the Apache config without Foreman installed.

Depends on https://github.com/theforeman/puppet-foreman/pull/800

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -181,6 +181,14 @@ class katello (
     wcache_page_size => $qpid_wcache_page_size,
   }
 
+  class { 'katello::application':
+    rest_client_timeout   => $rest_client_timeout,
+    cdn_ssl_version       => $cdn_ssl_version,
+    use_pulp_2_for_file   => $use_pulp_2_for_file,
+    use_pulp_2_for_docker => $use_pulp_2_for_docker,
+    repo_export_dir       => $repo_export_dir,
+  }
+
   class { 'katello::pulp':
     yum_max_speed            => $pulp_max_speed,
     num_workers              => $num_pulp_workers,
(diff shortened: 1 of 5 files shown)


ehelms:
The test failure look legitimate.

ekohl:
That looks like it ran before the PR was merged so I've restarted the tests.

ehelms:
Nah, its the same and valid failure. We dropped including `foreman` and there are references in `pulp.pp` to `foreman::user`.

ekohl:
And that's why I hate two services sharing a filesystem ;)

Just thinking out loud: we have this because Katello export uses this.

https://github.com/Katello/katello/blob/acaacd8bf7ac463ab8a35bf37ec555d185cc3754/app/models/setting/content.rb#L94-L95

This is probably in `/var/lib/pulp` because it copies files and that's faster if it's on the same filesystem.

https://github.com/Katello/katello/blob/53f50325a502997ced2239f2ecf11d1d31d45121/app/lib/actions/katello/repository/export.rb#L58-L59

How big would the impact be if we moved this somewhere where Foreman can write and to the file `application.pp`?

Perhaps a cleaner way is to move it to `application.pp` regardless and implement an `anchor` which can be collected with an `exec` that does `mkdir -p` so the parent directory actually exists.

ehelms:
Couple things to consider. First, Pulp does all the exporting so it has to be a location that Pulp (via the Apache user, although perhaps that should be the pulp user on newer installs) can write to. Hence the setup and location you see. We, as in Katello, when performing this sort of export then copy that Pulp export to a location of the users choosing. This is called the triple-disk-space problem with "legacy" Pulp exports.

Pulp 3 drops this export workflow in favor of a newer, core supported workflow.

My vote is to do the least amount of work to keep this workflow working as is since we will yank it all out after the Katello 3.16 release.

ekohl:
Let's see if this works.

manifests/init.pp
```diff
@@ -201,11 +208,4 @@
     repo_export_dir          => $repo_export_dir,
```
ehelms: I assume this needs to be dropped

ekohl:
An alternative is to drop this entire PR. Since the devel workflow doesn't use this class, it should have no effect there. If nightlies pass without this patch, we can drop this. Then after 3.16, it will be much simpler and the need is no longer there.

PR #325, opened 2020-03-03
Mark compatible with candlepin 9.x

ekohl:


ehelms:
Looks like the change to puma default broke a tests expectation on an exec

ehelms:
Can drop the second commit

ekohl:
Rebased

PR #332, opened 2020-04-09
Add postgresql-evr extension to katello

ianballou:
Adds the postgresql-evr package to katello.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -63,6 +63,8 @@ class katello::application (
   $candlepin_event_queue = $katello::params::candlepin_event_queue
   $crane_url = $katello::params::crane_url
   $crane_ca_cert = $certs::katello_server_ca_cert
+  $postgresql_evr_package = $katello::params::postgresql_evr_package
+  $manage_db = $foreman::db_manage
 
   # Katello database seeding needs candlepin
   Anchor <| title == 'katello::repo' or title ==  'katello::candlepin' |> ->
@@ -70,6 +72,12 @@ class katello::application (
     package     => $foreman::plugin_prefix.regsubst(/foreman_/, 'katello'),
     config      => template('katello/katello.yaml.erb'),
     config_file => "${foreman::plugin_config_dir}/katello.yaml",
+  }
+
+  if $manage_db {
+    package { $postgresql_evr_package:
+      ensure => installed,
+    }
   }
 
   foreman::config::apache::fragment { 'katello':
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -22,6 +22,8 @@
 #   The oauth key for Candlepin
 # @param candlepin_oauth_secret
 #   The oauth secret for Candlepin
+# @param postgresql_evr_package
+#   The contextual package name for the PostgreSQL EVR extension
 class katello::params (
   Stdlib::HTTPSUrl $pulp_url = "https://${facts['networking']['fqdn']}/pulp/api/v2/",
   Stdlib::HTTPSUrl $crane_url = "https://${facts['networking']['fqdn']}:5000",
@@ -29,6 +31,7 @@ class katello::params (
   String[1] $candlepin_oauth_key = $katello::globals::candlepin_oauth_key,
   String[1] $candlepin_oauth_secret = $katello::globals::candlepin_oauth_secret,
   Stdlib::Host $candlepin_host = 'localhost',
-  Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:8443/candlepin"
+  Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:8443/candlepin",
+  String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
 ) inherits katello::globals {
 }
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -32,4 +32,10 @@ class katello::globals(
 
   $candlepin_qpid_exchange = 'event'
   $candlepin_event_queue = 'katello_event_queue'
+
+  if $facts['os']['release']['major'] == '7' {
+    $postgresql_evr_package = 'rh-postgresql12-postgresql-evr'
+  } else {
+    $postgresql_evr_package = 'postgresql-evr'
+  }
 }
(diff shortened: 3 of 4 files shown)


manifests/application.pp
```diff
@@ -72,6 +73,10 @@
     config_file => "${foreman::plugin_config_dir}/katello.yaml",
   }
 
+  package { "${$scl_postgresql}-postgresql-evr":
```
ekohl, line 76: You don't need this if it's a remote database. This also breaks on EL8. You should really pass a variable `$postgresql_evr_package` and determine the full name in `params.pp` where you consider the OS version.

ianballou: This package is only build for EL7, so it looks like I will just do nothing if we're on EL8

ianballou:
@ekohl I think I've addressed your comments.

ianballou:
We have to bump the postgresql-evr version again.  I will update the version in this PR once that is up.

manifests/params.pp
```diff
@@ -22,13 +22,19 @@
...
   Stdlib::HTTPSUrl $pulp_url = "https://${facts['fqdn']}/pulp/api/v2/",
   Stdlib::HTTPSUrl $crane_url = "https://${facts['fqdn']}:5000",
   Stdlib::Host $qpid_hostname = 'localhost',
   String[1] $candlepin_oauth_key = $katello::globals::candlepin_oauth_key,
   String[1] $candlepin_oauth_secret = $katello::globals::candlepin_oauth_secret,
   Stdlib::Host $candlepin_host = 'localhost',
-  Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:8443/candlepin"
+  Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:8443/candlepin",
+  String[1] $scl_postgresql = 'rh-postgresql12',
+  String[1] $postgresql_evr_package = "postgresql-evr-0.0.2-1.el${$::os['release']['major']}.x86_64"
```
jlsherrill, line 38: i think we'd want just the package name here, and then use  :ensure => latest  in the package delcaration above (unless i'm completely off base here)

ianballou: Can I have the shortened package name here and still pick the right release (el7 vs el8)?

jlsherrill: yeah, its basically doing a 'yum install $PKG_NAME'  so as long as its doing   'rh-postgresql12-postgresql-evr' on el7 and 'postgresql-evr' on el8, i think we're good

ianballou: makes sense, I'll update

jlsherrill:
@ianballou do those failures make any sense to you?

ianballou:
@jlsherrill the don't, I'm going to test again locally.

manifests/application.pp
```diff
@@ -72,6 +75,12 @@
     config_file => "${foreman::plugin_config_dir}/katello.yaml",
   }
 
+  if $manage_db {
+    package { "${$scl_postgresql}-${postgresql_evr_package}":
```
ekohl, line 79: This doesn't allow no SCL as we want on EL8

manifests/application.pp
```diff
@@ -72,6 +75,12 @@
...
+      ensure => latest,
```
ekohl, line 80: We always use `installed`. Upgrading via Puppet is generally not what you want.
```suggestion
      ensure => installed,
```

manifests/params.pp
```diff
@@ -22,13 +22,19 @@
...
   Stdlib::HTTPSUrl $pulp_url = "https://${facts['fqdn']}/pulp/api/v2/",
   Stdlib::HTTPSUrl $crane_url = "https://${facts['fqdn']}:5000",
   Stdlib::Host $qpid_hostname = 'localhost',
   String[1] $candlepin_oauth_key = $katello::globals::candlepin_oauth_key,
   String[1] $candlepin_oauth_secret = $katello::globals::candlepin_oauth_secret,
   Stdlib::Host $candlepin_host = 'localhost',
-  Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:8443/candlepin"
+  Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:8443/candlepin",
+  String[1] $scl_postgresql = 'rh-postgresql12',
+  String[1] $postgresql_evr_package = 'postgresql-evr'
```
ekohl, line 38: What you can do is the following in `globals.pp`
```puppet
if $facts['os']['release']['major'] == '7' {
  $postgresql_evr_package = 'rh-postgresql12-postgresql-evr'
} else {
  $postgresql_evr_package = 'postgresql-evr'
}
```

Then:
```suggestion
  String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
```

This gives you sane defaults and ready for EL8 but allows you to override the package name if you need to and you want to experiment (via custom-hiera for example). If we change the SCL, we'd do a major version bump and change the hardcoded default.

ianballou:
I'm not sure why those warnings are still there, I'm not hitting that locally.

manifests/application.pp
```diff
@@ -63,6 +63,9 @@
   $candlepin_event_queue = $katello::params::candlepin_event_queue
   $crane_url = $katello::params::crane_url
   $crane_ca_cert = $certs::katello_server_ca_cert
+  $scl_postgresql = $katello::params::scl_postgresql
+  $postgresql_evr_package = $katello::params::postgresql_evr_package
+  $manage_db = katello::candlepin::manage_db
```
ekohl, line 68: This is incorrect. Katello and Foreman are separate databases.
```suggestion
  $manage_db = $foreman::db_manage
```

ekohl:
https://github.com/theforeman/puppet-katello/commit/6f42d22e45d34ef1615f59aef4d0e2b565510b9f was merged but had a bug and didn't actually ran the tests. That was fixed in a voxpupuli-tests 1.2 release, but I'm going through the modules to correct it, but that takes a bit of time.

ianballou:
@ekohl thanks for the review, I've pushed the changes.  Just waiting now to test them out.

jlsherrill:
[test]

jlsherrill:
this still isn't running, @ianballou maybe it needs a rebase?

manifests/application.pp
```diff
@@ -72,6 +74,12 @@
     config_file => "${foreman::plugin_config_dir}/katello.yaml",
   }
 
+  if $manage_db {
+    package { "${postgresql_evr_package}":
```
ehelms, line 78: No need for quotes here you can reference the variable

ianballou:
Tests are passing now, is the PR looking all set?

PR #335, opened 2020-04-23
Refs #29215: Only install Pulp and Qpid on EL7

ehelms:


manifests/application.pp
```diff
@@ -24,11 +24,14 @@
...
   include katello::params
 
-  include katello::qpid_client
-  User<|title == $foreman::user|>{groups +> 'qpidd'}
+
+  if $facts['os']['release']['major'] == '7' {
+    include certs::qpid
+    include katello::qpid_client
+    User<|title == $foreman::user|>{groups +> 'qpidd'}
+  }
```
ekohl, line 34: Does katello still connect to qpid? Can't we drop this altogether or move it to `pulp.pp`?

manifests/init.pp
```diff
@@ -173,34 +173,36 @@
...
-    mongodb_replica_set      => $pulp_db_replica_set,
-    mongodb_ssl              => $pulp_db_ssl,
-    mongodb_ssl_keyfile      => $pulp_db_ssl_keyfile,
-    mongodb_ssl_certfile     => $pulp_db_ssl_certfile,
-    mongodb_verify_ssl       => $pulp_db_verify_ssl,
-    mongodb_ca_path          => $pulp_db_ca_path,
-    mongodb_unsafe_autoretry => $pulp_db_unsafe_autoretry,
-    mongodb_write_concern    => $pulp_db_write_concern,
-    manage_mongodb           => $pulp_manage_db,
+  if $facts['os']['release']['major'] == '7' {
```
ekohl, line 183: I'm happy I separated these classes :)

manifests/application.pp
ekohl: To keep tracking: https://github.com/theforeman/puppet-katello/pull/336

manifests/application.pp
```diff
@@ -24,11 +24,14 @@
...
 
   foreman_config_entry { 'pulp_client_cert':
```
ekohl, line 36: Is this Pulp 2 only or both Pulp 2 & 3? If 2-only, this IMHO also needs conditionals, as well as the inclusion of `certs::pulp_client`.

manifests/application.pp
ehelms: I've included these updates in #336 so this PR now depends on those changes to go forward.

manifests/application.pp
ehelms: This is both.

ehelms:
This is stand alone now from https://github.com/theforeman/puppet-katello/pull/336 to get this change in and close out EL8 for this module.

mmoll:
Tests need updating.

ehelms:
I was like why did this not fail locally but I added a new supported version to the metadata and forgot to re-run.

ehelms:
Including it in the metadata was a good update as it brought lots of identification of areas in the tests that need scoping of expectations based on major release number. That is fixed now.

spec/classes/application_spec.rb
```diff
@@ -16,11 +16,24 @@ class { 'katello::params':
 
       context 'with default parameters' do
         it { is_expected.to compile.with_all_deps }
-        it { is_expected.to create_package('tfm-rubygem-katello') }
-        it { is_expected.not_to create_package('tfm-rubygem-katello').that_requires('Anchor[katello::candlepin]') }
-        it { is_expected.to create_package('rh-postgresql12-postgresql-evr') }
-        it { is_expected.to contain_class('certs::qpid') }
-        it { is_expected.to contain_class('katello::qpid_client') }
+
+        if facts[:operatingsystemmajrelease] == '7'
```
ehelms, line 20: Is this the right way to scope the testing? I saw it used in other modules so I borrowed it throughout. @mmoll @ekohl

mmoll: In general yes, however in this specific case (if you have the time), I feel having something like this will be a tad better:
(pseudo-code)
```
  rubygem_prefix = 'tfm-'
  postgres_prefix = 'rh-postgresql12-'
else
  rubygem_prefix = ''
  postgres_prefix = ''
end
[...]
          it { is_expected.to create_package("#{rubygem_prefix}rubygem-katello") }
          it { is_expected.not_to create_package("#{rubygem_prefix}rubygem-katello").that_requires('Anchor[katello::candlepin]') }
          it { is_expected.to create_package("{postgres_prefix}postgresql-evr") }
```

ehelms: Given this will get ripped out in less than 3 months, would you still want this level of generic support?

mmoll: OK, nope :)

mmoll:
merged, thanks @ehelms!

wbclark:
Changed the title of this PR to make it more useful in changelogs

PR #336, opened 2020-04-27
Refs #28924: Drop qpid client configuration used by older katello-qpid

ehelms:
In theory we can drop this, giving it a test first.


ehelms:
Cleared tests

ehelms:
Devel module updates:

https://github.com/theforeman/puppet-katello_devel/pull/236

manifests/pulp.pp
```diff
@@ -84,6 +84,7 @@
 ) {
   include katello::params
   include certs
+  include certs::qpid
 
   class { 'certs::qpid_client':
```
ekohl, line 89: You deleted this class but still include it here.

ehelms: I dropped `katello::qpid_client`. This one handles client certificates for Pulp effectively to talk to Qpid.

ekohl: I shouldn't do reviews when I'm not fully awake ... I assumed the test failure was due to this

ehelms:
Finally got around to updating the tests.

manifests/pulp.pp
```diff
@@ -100,6 +100,7 @@
 ) {
   include katello::params
   include certs
+  include certs::qpid
```
ekohl, line 103: This feels odd to me. Why does Pulp need qpid server certificates? However, the same is done [here](https://github.com/theforeman/puppet-foreman_proxy_content/blob/34aa29ab3acd98b4fc403c0c778e0c67f7d9b53c/manifests/init.pp#L279-L282). I don't think it's worth to clean that up given Pulp 2 will be going away.

@timogoebel @laugmanuel do you think this will be an issue in your split deployments?

ehelms: Suppose Pulp itself does not need them, but they are only relevant when Pulp is involved so we need to ensure they get setup if Pulp is present for it's tasking system.

ekohl: I always thought `certs::qpid` were there to run a qpid server while `certs::qpid_client` were there to connect to a qpid server. Pulp only connects to a pulp server, so shouldn't this use `certs::qpid_client`?

ehelms: That's true. So the confusing part is then why they were where they were in the first place. I'll update and drop this from here.

PR #337, opened 2020-05-04
Support manage_httpd to allow Pulp in a standalone setup (followup)

laugmanuel:
Follow up to #312

This PR enables Pulp to exist as a standalone system and therefore enables split deployments.

The default behaviour does not change in any way. It it explicitly required to set a least manage_httpd to true.

Note: The original PR is #312. Because the source branch was missing, I needed to open a new PR with an existing branch I can push to.

### DIFF
--- a/manifests/pulp.pp
+++ b/manifests/pulp.pp
@@ -63,8 +63,8 @@
 # @param pub_dir_options
 #   The Apache options to use on the `/pub` resource
 #
-# @param manage_httpd
-#   Boolean to also manage httpd for Pulp in a standalone setup
+# @param manage_vhost_standalone
+#   Boolean to manage the Pulp vhost standalone. Otherwise the vhost is managed as part of Foreman
 #
 # @param https_cert
 #   The Apache public certificate for ssl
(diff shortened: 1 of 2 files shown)


laugmanuel:
@ekohl sure we can rename the parameter. What do you think about `manage_httpd_standalone` or something like that? This would also indicate the change in behaviour.

ekohl:
You could say that `pulp` is standalone, but it's more about a vhost than `httpd` (yes, badly named in puppet-pulp) so `Boolean $manage_vhost_standalone`? However, I don't really know what it would do if you say false so I was thinking in the direction of `Enum['foreman', 'pulp'] $vhost`. It's complicated.

For now I'm thinking the easiest would be `Boolean $manage_vhost_standalone` and document that it's part of the Foreman vhost when it's not standalone in the parameter documentation. What do you think?

laugmanuel:
Sounds good to me. I changed the name of the variable and improved the parameter documentation accordingly.

ekohl:
(I'll let the tests continue, but the code looks good to merge).

PR #340, opened 2020-05-11
Fixes #29715: Add option to expose mongodb on a public interface with…

ehelms:
… SSL

Requires:

 * https://github.com/theforeman/puppet-pulp/pull/400
 * https://github.com/theforeman/puppet-certs/pull/285

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -91,6 +91,11 @@
 #
 # $pulp_manage_db::      Boolean to install and configure the mongodb.
 #
+# $pulp_expose_db::        Expose Pulp database on a public interface for Pulp 3 migrations
+#
+# $pulp_expose_db_ipv6::   When exposing the Pulp database on a public interface, also listen on IPv6.
+#                          Ignored when pulp_expose_db is false.
+#
 # $use_pulp_2_for_file::   Configures Katello to use Pulp 2 for file content
 #
 # $use_pulp_2_for_docker:: Configures Katello to use Pulp 2 for docker content
@@ -144,6 +149,8 @@ class katello (
   Boolean $pulp_db_unsafe_autoretry = false,
   Optional[Enum['majority', 'all']] $pulp_db_write_concern = undef,
   Boolean $pulp_manage_db = true,
+  Boolean $pulp_expose_db = false,
+  Boolean $pulp_expose_db_ipv6 = true,
 ) {
 
   package { 'katello':
@@ -207,6 +214,8 @@ class katello (
       mongodb_unsafe_autoretry => $pulp_db_unsafe_autoretry,
       mongodb_write_concern    => $pulp_db_write_concern,
       manage_mongodb           => $pulp_manage_db,
+      expose_mongodb           => $pulp_expose_db,
+      expose_mongodb_ipv6      => $pulp_expose_db_ipv6,
     }
   }
 }
--- a/manifests/pulp.pp
+++ b/manifests/pulp.pp
@@ -75,6 +75,13 @@
 # @param https_ca_cert
 #   The Apache CA certificate for client authentication
 #
+# @param expose_mongodb
+#   Boolean to expose MongoDB server on a public interface
+#
+# @param expose_mongodb_ipv6
+#   When exposing the Pulp database on a public interface, also listen on IPv6.
+#   Ignored when pulp_expose_db is false.
+#
 class katello::pulp (
   Optional[String] $yum_max_speed = undef,
   Optional[Integer[1]] $num_workers = undef,
@@ -97,9 +104,15 @@ class katello::pulp (
   Optional[Stdlib::Absolutepath] $https_cert = undef,
   Optional[Stdlib::Absolutepath] $https_key = undef,
   Optional[Stdlib::Absolutepath] $https_ca_cert = undef,
+  Boolean $expose_mongodb = false,
+  Boolean $expose_mongodb_ipv6 = true,
 ) {
   include katello::params
   include certs
+
+  if !manage_mongodb and $expose_mongodb {
+    fail('Exposing MongoDB is only possible on managed databases. Either manage the database or do not expose.')
+  }
 
   class { 'certs::qpid_client':
     require => Class['pulp::install'],
@@ -127,6 +140,52 @@ class katello::pulp (
       content     => template('katello/pulp-apache.conf.erb'),
       ssl_content => template('katello/pulp-apache-ssl.conf.erb'),
     }
+  }
+
+  if $expose_mongodb {
+    include certs::mongodb
+    include certs::mongodb_client
+
+    Class['certs::mongodb_client'] ~> Class['pulp', 'apache::service']
+
+    $mongodb_bind_ip = $expose_mongodb_ipv6 ? {
+      true  => ['0.0.0.0', '::0'],
+      false => ['0.0.0.0'],
+    }
+
+    class { 'mongodb::server':
+      bind_ip => $mongodb_bind_ip,
+      ipv6    => $expose_mongodb_ipv6,
+      ssl     => true,
+      ssl_key => $certs::mongodb::mongodb_server_bundle,
+      ssl_ca  => $certs::mongodb::mongodb_server_ca_cert,
+      require => Class['certs::mongodb'],
+    }
+
+    $mongodb_seeds_real = "${certs::mongodb::hostname}:27017"
+    $db_password = extlib::cache_data('foreman_cache_data', 'mongo_db_password', extlib::random_password(32))
+    $mongo_cmd = "mongo admin --quiet --host ${mongodb_seeds_real} --ssl --sslPEMKeyFile ${certs::mongodb_client::mongodb_client_bundle} --sslCAFile ${certs::mongodb_client::mongodb_client_ca_cert}"
+
+    exec { 'mongodb user':
+      command   => "${mongo_cmd} --eval \"db.createUser({user:'pulp',pwd:'${db_password}',roles:[{role:'dbOwner', db:'${mongodb_name}'},{ role: 'readWrite', db: '${mongodb_name}'}]})\"",
+      path      => '/usr/bin:/bin',
+      logoutput => 'on_failure',
+      unless    => "${mongo_cmd} --eval 'printjson(db.system.users.find({\"_id\":\"${mongodb_name}.pulp\"}).toArray().length)' | grep -q 1",
+      require   => CLass['mongodb::server'],
+    }
+
+    $mongodb_ssl_keyfile_real = $certs::mongodb_client::mongodb_client_key
+    $mongodb_ssl_certfile_real = $certs::mongodb_client::mongodb_client_cert
+    $mongodb_ca_path_real = $certs::mongodb_client::mongodb_client_ca_cert
+    $mongodb_ssl_real = true
+    $db_username_real = 'pulp'
+    $db_password_real = $db_password
+  } else {
+    $mongodb_seeds_real = $mongodb_seeds
+    $mongodb_ssl_keyfile_real = $mongodb_ssl_keyfile
+    $mongodb_ssl_certfile_real = $mongodb_ssl_certfile
+    $mongodb_ca_path_real = $mongodb_ca_path
+    $mongodb_ssl_real = $mongodb_ssl
   }
 
   Anchor <| title == 'katello::repo' |> -> # lint:ignore:anchor_resource
@@ -161,15 +220,15 @@ class katello::pulp (
     subscribe              => Class['certs'],
     worker_timeout         => $worker_timeout,
     db_name                => $mongodb_name,
-    db_seeds               => $mongodb_seeds,
+    db_seeds               => $mongodb_seeds_real,
     db_username            => $mongodb_username,
     db_password            => $mongodb_password,
     db_replica_set         => $mongodb_replica_set,
-    db_ssl                 => $mongodb_ssl,
-    db_ssl_keyfile         => $mongodb_ssl_keyfile,
-    db_ssl_certfile        => $mongodb_ssl_certfile,
+    db_ssl                 => $mongodb_ssl_real,
+    db_ssl_keyfile         => $mongodb_ssl_keyfile_real,
+    db_ssl_certfile        => $mongodb_ssl_certfile_real,
     db_verify_ssl          => $mongodb_verify_ssl,
-    db_ca_path             => $mongodb_ca_path,
+    db_ca_path             => $mongodb_ca_path_real,
     db_unsafe_autoretry    => $mongodb_unsafe_autoretry,
     db_write_concern       => $mongodb_write_concern,
     manage_db              => $manage_mongodb,
--- a/spec/classes/pulp_spec.rb
+++ b/spec/classes/pulp_spec.rb
@@ -92,6 +92,57 @@ describe 'katello::pulp' do
               end
             end
           end
+
+          context 'with expose_mongodb parameter' do
+            let :params do
+              {
+                expose_mongodb: true,
+              }
+            end
+
+            it do
+              is_expected.to compile.with_all_deps
+              is_expected.to contain_class('certs::mongodb')
+              is_expected.to contain_class('certs::mongodb_client')
+
+              is_expected.to create_class('mongodb::server')
+                .with_bind_ip(['0.0.0.0', '::0'])
+                .with_ipv6(true)
+                .with_ssl(true)
+                .with_ssl_key('/etc/pki/katello/mongodb/mongodb-server-certificate-bundle.pem')
+                .with_ssl_ca('/etc/pki/katello/certs/katello-server-ca.crt')
+
+              is_expected.to create_class('pulp')
+                .with_db_seeds("#{facts[:fqdn]}:27017")
+                .with_db_ssl_keyfile('/etc/pulp/mongodb/mongodb-client-certificate.key')
+                .with_db_ssl_certfile('/etc/pulp/mongodb/mongodb-client-certificate.crt')
+                .with_db_ca_path('/etc/pki/katello/certs/katello-server-ca.crt')
+            end
+
+            context 'with mongodb_ipv6 => false' do
+              let :params do
+                super().merge({ 'expose_mongodb_ipv6' => false, })
+              end
+
+              it do
+                is_expected.to compile.with_all_deps
+                is_expected.to contain_class('certs::mongodb')
+
+                is_expected.to create_class('mongodb::server')
+                  .with_bind_ip(['0.0.0.0'])
+                  .with_ipv6(false)
+                  .with_ssl(true)
+                  .with_ssl_key('/etc/pki/katello/mongodb/mongodb-server-certificate-bundle.pem')
+                  .with_ssl_ca('/etc/pki/katello/certs/katello-server-ca.crt')
+
+                is_expected.to create_class('pulp')
+                  .with_db_seeds("#{facts[:fqdn]}:27017")
+                  .with_db_ssl_keyfile('/etc/pulp/mongodb/mongodb-client-certificate.key')
+                  .with_db_ssl_certfile('/etc/pulp/mongodb/mongodb-client-certificate.crt')
+                  .with_db_ca_path('/etc/pki/katello/certs/katello-server-ca.crt')
+              end
+            end
+          end
         end
       end
     end


manifests/pulp.pp
```diff
@@ -129,50 +134,85 @@
...
+    db_verify_ssl             => $mongodb_verify_ssl,
+    db_ca_path                => $mongodb_ca_path_real,
+    db_unsafe_autoretry       => $mongodb_unsafe_autoretry,
+    db_write_concern          => $mongodb_write_concern,
+    manage_db                 => $manage_mongodb,
+    mongodb_bind_ip           => $mongodb_bind_ip,
+    mongodb_server_ssl        => $mongodb_server_ssl,
+    mongodb_server_ssl_bundle => $mongodb_server_ssl_bundle,
+    mongodb_server_ssl_ca     => $mongodb_server_ssl_ca,
+    mongodb_ipv6              => $expose_mongodb_ipv6,
```
ekohl, line 215: Rather than passing the parameters, I'd prefer to explicitly define `mongodb::server` just before the `pulp` class.

ehelms: I'm going to need some help understanding *why* that design choice is better. And, probably then, why we have an `include mongodb::server` in the pulp module at all. I have a hard time following the design granularities and when to use what.

ekohl: The reasoning is that you can only manage a class once. Both the MongoDB and PostgreSQL services are classes. In regular programming languages you'd refer to them as singletons. It's not possible to instantiate multiple instances. If multiple services depend on a service (think both Foreman and Pulpcore that need PostgreSQL) then it's a conflict. That's why Foreman only includes PostgreSQL and ensures the correct chaining. (first DB, then DB migrations). The same thing with Pulpcore. `include` is what a `.getInstance()` would be in other languages where `class myclass {}` is like calling the constructor.

To managing it, there are 2 solutions: hiera (what we in the installer do for PostgreSQL) or calling the constructor before ever including it. Here you'd take the latter approach.

Typically this is reserved for profile modules which combine multiple basic modules into a specific deployment. I could puppet-katello as a profile module.

manifests/pulp.pp
```diff
@@ -129,50 +134,79 @@
...
-    db_replica_set         => $mongodb_replica_set,
-    db_ssl                 => $mongodb_ssl,
-    db_ssl_keyfile         => $mongodb_ssl_keyfile,
-    db_ssl_certfile        => $mongodb_ssl_certfile,
-    db_verify_ssl          => $mongodb_verify_ssl,
-    db_ca_path             => $mongodb_ca_path,
-    db_unsafe_autoretry    => $mongodb_unsafe_autoretry,
-    db_write_concern       => $mongodb_write_concern,
-    manage_db              => $manage_mongodb,
+    server_name               => $server_name,
```
ekohl, line 168: Mind changing the whitespace back?

manifests/pulp.pp
```diff
@@ -129,50 +134,79 @@
...
+    puppet_wsgi_processes     => 1,
+    enable_katello            => true,
+    subscribe                 => Class['certs'],
+    worker_timeout            => $worker_timeout,
+    db_name                   => $mongodb_name,
+    db_seeds                  => $mongodb_seeds_real,
+    db_username               => $mongodb_username,
+    db_password               => $mongodb_password,
+    db_replica_set            => $mongodb_replica_set,
+    db_ssl                    => $mongodb_ssl or $mongodb_server_ssl,
```
ekohl, line 202: Looks like ` or $mongodb_server_ssl` can be dropped.

manifests/pulp.pp
```diff
@@ -129,50 +134,79 @@
     }
   }
 
+  if $expose_mongodb {
```
ekohl, line 137:
```suggestion
  if $manage_mongodb and $expose_mongodb {
```

manifests/pulp.pp
```diff
@@ -129,50 +134,79 @@
...
+    include certs::mongodb
+
+    Class['certs::mongodb'] -> Class['mongodb::server']
```
ekohl, line 140: You can also add a `require => Class['certs::mongodb']` on the `class { 'mongodb::server':` definition`.

manifests/pulp.pp
ehelms: Can these two lines (https://github.com/theforeman/puppet-pulp/blob/master/manifests/database.pp#L7-L8) be defined with the `include mongodb::server` ?

ekohl: No, `include` can't have a require statement. `include` really means "ensure it's somewhere in the catalog" but makes no guarantees whatsoever about ordering. In that particular case, I think `require mongodb::server` would be the statement you're looking for.

manifests/pulp.pp
```diff
@@ -129,50 +134,81 @@
     }
   }
 
+  if $expose_mongodb {
+    include certs::mongodb
+    include certs::mongodb_client
```
ekohl, line 139: I think this should be unconditional now: we always need the client certs so moved out of this if block.

manifests/pulp.pp
ehelms: Expose mongodb should fail if manage_mongodb is false... should I go the explicit fail route instead?

manifests/pulp.pp
ehelms: We only need the mongodb client certificates if we are exposing mongodb, otherwise they go unused.

manifests/pulp.pp
ekohl: I think that makes sense.

manifests/pulp.pp
ekohl: Right. And we'll also need it in foreman_proxy_content with pulpcore migrator plugin.

manifests/init.pp
```diff
@@ -91,6 +91,10 @@
 #
 # $pulp_manage_db::      Boolean to install and configure the mongodb.
 #
+# $pulp_expose_db::        Expose Pulp database on a public interface for Pulp 3 migrations
+#
+# $pulp_expose_db_ipv6::   Expose Pulp database on IPv6 public interface for Pulp 3 migrations
```
ekohl, line 96: How about something like:
```suggestion
# $pulp_expose_db_ipv6::   When exposing the Pulp database on a public interface, also listen on IPv6. Ignored when pulp_expose_db is false.
```

manifests/init.pp
```diff
@@ -141,8 +145,14 @@
   Boolean $pulp_db_unsafe_autoretry = false,
   Optional[Enum['majority', 'all']] $pulp_db_write_concern = undef,
   Boolean $pulp_manage_db = true,
+  Boolean $pulp_expose_db = false,
+  Boolean $pulp_expose_db_ipv6 = false,
 ) {
 
+  if !$pulp_manage_db and $pulp_expose_db {
+    fail("pulp_expose_db cannot be set to true if pulp_manage_db is false")
+  }
```
ekohl, line 154: I think it makes more sense to have this in `pulp.pp`. It's compiled anyway so it fails regardless, but that way the standalone use of `katello::pulp` also benefits from the check. Design wise it's also better to keep all Pulp logic in Pulp itself. When dropping Pulp 2, we can simply remove pulp.pp and drop all unused parameters.

manifests/pulp.pp
```diff
@@ -97,10 +100,20 @@
...
   include katello::params
   include certs
 
+  if !$pulp_manage_db and $pulp_expose_db {
+    fail("pulp_expose_db cannot be set to true if pulp_manage_db is false")
+  }
+
+  if !manage_mongodb and $expose_mongodb {
+    fail("expose_mongodb cannot be set to true if manage_mongodb is false")
+  }
```
ehelms, line 115: @ekohl Like this? This feels a bit weird having to define it twice due to having two sets of parameters.

ekohl: Just defining it once is enough.

ehelms: The way I was looking at it,  the parameters are different. Therefore the messaging will need to be different depending on if your entry point is `init.pp` vs `katello::pulp`. Otherwise the message could be confusing since the parameters have different names.

ekohl: I'd say the explicit parameters do not need to be mentioned. If you think about the installer context, it'll be confusing that you specified `--katello-expose-mongodb` but then the error talks about `expose_mongodb`. That's why I'd phrase it more generic. Something like:

Exposing MongoDB is only possible on managed databases. Either manage the database or do not expose.

spec/classes/pulp_spec.rb
```diff
@@ -92,6 +92,51 @@
...
+          context 'with expose_db parameter' do
+            let :params do
+              {
+                expose_mongodb: true,
+              }
+            end
+
+            it do
+              is_expected.to compile.with_all_deps
+              is_expexted.to contain_class('certs::mongodb')
```
wbclark, line 105:
```suggestion
              is_expected.to contain_class('certs::mongodb')
```

spec/classes/pulp_spec.rb
```diff
@@ -92,6 +92,51 @@
...
+            end
+
+            context 'with mongodb_ipv6 => true' do
+              let :params do
+                super().merge({ 'expose_mongodb_ipv6' => true, })
+              end
+
+              it do
+                is_expected.to compile.with_all_deps
+                is_expexted.to contain_class('certs::mongodb')
```
wbclark, line 125:
```suggestion
                is_expected.to contain_class('certs::mongodb')
```

manifests/pulp.pp
```diff
@@ -129,6 +142,39 @@
...
+      require => Class['certs::mongodb'],
+    }
+
+    $mongodb_seeds_real = "${facts['networking']['os']}:27017"
+    $mongodb_ssl_keyfile_real = $certs::mongodb_client::mongodb_client_key
+    $mongodb_ssl_certfile_real = $certs::mongodb_client::mongodb_client_cert
+    $mongodb_ca_path_real = $certs::mongodb_client::mongodb_client_ca_cert
+    $mongodb_ssl_real = true
+  } else {
+    $mongodb_seeds_real = 'localhost:27017'
```
wbclark, line 171:
```suggestion
    $mongodb_seeds_real = $mongodb_seeds
```
This is breaking the existing test where `mongodb_seeds` is overridden at spec/classes/pulp_spec.rb:71 .

manifests/pulp.pp
```diff
@@ -129,6 +142,39 @@
...
+  if $expose_mongodb {
+    include certs::mongodb
+    include certs::mongodb_client
+
+    Class['certs::mongodb_client'] -> Class['pulp']
+
+    $mongodb_bind_ip = $expose_mongodb_ipv6 ? {
+      true  => ['0.0.0.0', '::/0'],
+      false => ['0.0.0.0'],
+    }
```
wbclark, line 154:
```suggestion
    }

    $mongodb_ipv6 = $expose_mongodb_ipv6 ? {
      true  => true,
      false => undef,
    }
```

I think this would work.

ehelms: Why is this needed?

wbclark: It's not. I like your solution better.

ehelms:
When I test this through the full installer, where I run setting expose-db after an installation I am finding that Apache is not restarted after the `server.conf` gets updated such that the install is broken because Pulp cannot connect. I expected (https://github.com/theforeman/puppet-pulp/blob/master/manifests/init.pp#L572) to handle this.

What am I missing?

wbclark:
@ehelms I think I understand why that happens:

1. Nothing in https://github.com/theforeman/puppet-katello/blob/master/manifests/pulp.pp#L111-L130 notifies the httpd service, and even if it did....

2. $manage_vhost_standalone defaults to false so I think that block doesn't get executed in your case. What does happen is that the Pulp class is declared with $manage_httpd => false.

3. Therefore you don't get any of the apache::vhost declarations in https://github.com/theforeman/puppet-pulp/blob/master/manifests/apache.pp#L22-L127

4. What you do get is the following file declaration: https://github.com/theforeman/puppet-pulp/blob/master/manifests/apache.pp#L129-L137

Since #4 is just a file that is not tied to any service and doesn't notify any other classes, you aren't getting the service restart as needed.

wbclark:
Actually I missed something, there is still notify => Service['httpd'] at https://github.com/theforeman/puppet-pulp/blob/master/manifests/apache.pp#L135

wbclark:
So, I think the issue is a bit simpler than I previously thought:

When $expose_mongodb is changed to true in katello::pulp what happens? We don't make any changes to the actual pulp classes -- we're actually touching certs and mongodb.

manifests/pulp.pp
```diff
@@ -129,6 +142,39 @@
     }
   }
 
+  if $expose_mongodb {
+    include certs::mongodb
+    include certs::mongodb_client
+
+    Class['certs::mongodb_client'] -> Class['pulp']
```
wbclark, line 149:
```suggestion
    Class['certs::mongodb_client'] ~> Class['pulp']
```

I think this will fix your issue, @ehelms

ehelms: I thin kthat will only cover if the certificates get created or changed? But doesn't cover I feel the whole of the issue which is if `/etc/pulp/server.conf` gets updated, then `httpd` should get restarted to pick up the changes. 

@ekohl adding you in case you had any thoughts as to why I don't seem to be seeing the refreshing happen

spec/classes/pulp_spec.rb
```diff
@@ -92,6 +92,57 @@
...
+              is_expected.to create_class('pulp')
+                .with_db_seeds("#{facts[:fqdn]}:27017")
+                .with_db_ssl_keyfile('/etc/pulp/mongodb/mongodb-client-certificate.key')
+                .with_db_ssl_certfile('/etc/pulp/mongodb/mongodb-client-certificate.crt')
+                .with_db_ca_path('/etc/pki/katello/certs/katello-server-ca.crt')
+            end
+
+            context 'with mongodb_ipv6 => true' do
+              let :params do
+                super().merge({ 'expose_mongodb_ipv6' => true, })
```
ekohl, line 124: I've grown accustomed to the shorter syntax:
```suggestion
                super().merge(expose_mongodb_ipv6: true)
```

manifests/pulp.pp
ekohl: I suspect Pulp doesn't actually manage the vhost and thus doesn't refresh Apache. I'd try `Class['certs::mongodb_client'] ~> Class['pulp', 'apache::service']`. Both since `pulp` will still refresh the workers.

ehelms:
This will need -- https://github.com/theforeman/puppet-certs/pull/288

ehelms:
@wbclark  @ekohl ready for review now, not the prettiest but nothing with MongoDB ever is >.<

manifests/pulp.pp
```diff
@@ -129,6 +142,52 @@
...
+
+    exec { 'mongodb user':
+      command   => "${mongo_cmd} --eval \"db.createUser({user:'pulp',pwd:'${db_password}',roles:[{role:'dbOwner', db:'pulp_database'},{ role: 'readWrite', db: 'pulp_database'}]})\"",
+      path      => '/usr/bin:/bin',
+      logoutput => 'on_failure',
+      unless    => "${mongo_cmd} --eval 'printjson(db.system.users.find({\"_id\":\"pulp_database.pulp\"}).toArray().length)' | grep -q 1",
+      require   => CLass['mongodb::server'],
+    }
+
+    $mongodb_seeds_real = "${facts['networking']['fqdn']}:27017"
```
ekohl, line 176: This is essentially also used in $mongo_cmd. Could you define this earlier and reuse it there?

manifests/pulp.pp
```diff
@@ -129,6 +142,52 @@
...
+      bind_ip => $mongodb_bind_ip,
+      ipv6    => $expose_mongodb_ipv6,
+      ssl     => true,
+      ssl_key => $certs::mongodb::mongodb_server_bundle,
+      ssl_ca  => $certs::mongodb::mongodb_server_ca_cert,
+      require => Class['certs::mongodb'],
+    }
+
+    $db_password = extlib::cache_data('foreman_cache_data', 'mongo_db_password', extlib::random_password(32))
+    $mongo_cmd = "mongo admin --quiet --host ${facts['networking']['fqdn']}:27017 --ssl --sslPEMKeyFile ${certs::mongodb_client::mongodb_client_bundle} --sslCAFile ${certs::mongodb_client::mongodb_client_ca_cert}"
```
ekohl, line 166: You could use `$mongodb::server::port` instead of hardcoding it.

manifests/pulp.pp
```diff
@@ -129,6 +142,52 @@
...
+
+    exec { 'mongodb user':
+      command   => "${mongo_cmd} --eval \"db.createUser({user:'pulp',pwd:'${db_password}',roles:[{role:'dbOwner', db:'pulp_database'},{ role: 'readWrite', db: 'pulp_database'}]})\"",
```
ekohl, line 169: We have a variable for `pulp_database` with `$mongodb_name`, right?

manifests/pulp.pp
```diff
@@ -129,6 +142,52 @@
...
+      path      => '/usr/bin:/bin',
```
ekohl, line 170: Does this work with SCLs?

ehelms: Yep, there is a `/usr/bin/mongo` deployed by the `-syspaths` RPM that provides a proper wrapper.

manifests/pulp.pp
```diff
@@ -129,6 +142,53 @@
     }
   }
 
+  if $expose_mongodb {
+    include certs::mongodb
+    include certs::mongodb_client
+
+    Class['certs::mongodb_client'] ~> Class['pulp', 'apache::service']
+
+    $mongodb_seeds_real = "${facts['networking']['fqdn']}:27017"
```
ekohl, line 151:
```suggestion
    $mongodb_seeds_real = "${facts['networking']['fqdn']}:${mongodb::server::port}"
```

ehelms: They are 99.9% of the time going to be the same thing because this is a tightly controlled scenario. But I shall may the change.

ehelms: The use of `mongodb::server::port` is incorrect here since port is undef -- https://github.com/voxpupuli/puppet-mongodb/blob/master/manifests/server.pp#L27

ekohl: > They are 99.9% of the time going to be the same thing because this is a tightly controlled scenario. But I shall may the change.

The important part is showing that these are linked. That's what variables are for. It gives understanding to the reader of the code.

> The use of `mongodb::server::port` is incorrect here since port is undef -- https://github.com/voxpupuli/puppet-mongodb/blob/master/manifests/server.pp#L27

Ah right. Fair enough.

manifests/pulp.pp
```diff
@@ -97,10 +104,16 @@
   Optional[Stdlib::Absolutepath] $https_cert = undef,
   Optional[Stdlib::Absolutepath] $https_key = undef,
   Optional[Stdlib::Absolutepath] $https_ca_cert = undef,
+  Boolean $expose_mongodb = false,
+  Boolean $expose_mongodb_ipv6 = false,
```
ekohl, line 108: Why would this default to false? In which case are you not able to bind to IPv6? Note that in our install Redis already binds to `::1` so not having IPv6 is already a requirement. And if you go that far, why do we need a parameter for it?

ehelms: I think you are saying we can default to exposing on IPv4 and IPv6 in all cases? Is there a case where IPv4 only is required? Or IPv6 only is required?

ekohl: Indeed I am. I think IPv4-only would be required if the kernel has no IPv6 support but we don't support that case anymore. I'm not sure you can actually run the Linux kernel without IPv4 support so binding should always work.

Maybe I'm missing something, but I'd prefer to at least default to true. Another option is:
```puppet
  Variant[Boolean, Array[Stdlib::IP::Address]] $expose_mongodb = false,
```

Then `true` would imply dual stack. If the user needs full control over exact binding, you have it and only need a single parameter.

ehelms: I set the default to true and left it there given this is designed for a specific use case and I think the parameters provide the needed flexibility.

manifests/pulp.pp
```diff
@@ -129,6 +142,52 @@
...
+      bind_ip => $mongodb_bind_ip,
+      ipv6    => $expose_mongodb_ipv6,
+      ssl     => true,
+      ssl_key => $certs::mongodb::mongodb_server_bundle,
+      ssl_ca  => $certs::mongodb::mongodb_server_ca_cert,
+      require => Class['certs::mongodb'],
+    }
+
+    $mongodb_seeds_real = "${certs::mongodb::hostname}:27017"
+    $db_password = extlib::cache_data('foreman_cache_data', 'mongo_db_password', extlib::random_password(32))
```
ekohl, line 166: We already have `$mongodb_password` as a parameter and IMH we should at least prefer that if specified.

ehelms: That password is intended for the use case when a user is providing credentials for an external database. I don't think it should apply to this workflow.

ekohl: I think it does. You're exposing the user with a certain password. The parameter allows the end user to specify something. That makes it easier on the consuming node to use the same password. Now the end user needs to look in files to see which password was generated.

More concrete: we have `server-1.example.com` which runs Katello with Pulp and MongoDB. The user runs `foreman-installer --katello-pulp-mongodb-password mysecretpassword`.

Now the user wants to consume this on `server-2.example.com`. The user runs `foreman-installer --foreman-proxy-content-mongo-db-password mysecretpassword`.

The current implementation doesn't support this because it's silently ignored.

ehelms: My point, unless I am just missing something is, that nobody should be doing this other than our own tooling which is tightly controlled for a very specific upgrade scenario.

ekohl: So how do you get the password now? Do you parse it out of a config file later on?

ehelms: That will be the most likely tactic given it's the only reliable way to get the value. I think I get what you are saying now but not form the users perspective, but the *if* the parameter is used then it can be stored on the answers file. But I don't see how you set the parameter conditionally.

PR #341, opened 2020-05-13
Add postgresql-debversion extension to katello

m-bucher:
educated copy of https://github.com/theforeman/puppet-katello/pull/332 for the `debversion`-extension

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -59,6 +59,7 @@ class katello::application (
   $candlepin_events_ssl_key = $certs::candlepin::client_key
   $crane_url = $katello::params::crane_url
   $crane_ca_cert = $certs::katello_server_ca_cert
+  $postgresql_debversion_package = $katello::params::postgresql_debversion_package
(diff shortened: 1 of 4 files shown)


manifests/application.pp
```diff
@@ -71,6 +72,9 @@
   }
 
   if $manage_db {
+    package { $postgresql_debversion_package:
+      ensure => installed,
+    }
```
ekohl, line 77: I think this should only be done if `$enable_deb` is true, right? Or will it be needed unconditionally?

m-bucher: I think we need it unconditionally, because the database-migration in katello will also run unconditionally.

jlsherrill: yes, whether a content type is enabled/installed doesn't change the schema that is deployed, so its sitll needed

ekohl:
Revert PR: https://github.com/theforeman/puppet-katello/pull/353

PR #346, opened 2020-05-26
Fixes #29944 - Make hosts_queue number of workers configurable

ianballou:
Adds support for configuring the number of Dynflow workers for the hosts_queue.

The idea here is to add unique IDs to the `hosts_queue_worker_ids` array to generate more Dynflow workers.

This PR will require a change to foreman-maintain to restart all of the running `worker-hosts-queue` workers.

Questions:
Adding more workers with an array doesn't seem natural to me, but Puppet seems to have limited looping capabilities.  Is there a better way to do this with Puppet?

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -66,6 +66,7 @@ class katello::application (
   $postgresql_debversion_package = $katello::params::postgresql_debversion_package
   $postgresql_evr_package = $katello::params::postgresql_evr_package
   $manage_db = $foreman::db_manage
+  $hosts_queue_workers = $katello::params::hosts_queue_workers
 
   # Katello database seeding needs candlepin
   Anchor <| title == 'katello::repo' or title ==  'katello::candlepin' |> ->
@@ -89,8 +90,10 @@ class katello::application (
   }
 
   if $foreman::jobs_manage_service {
-    foreman::dynflow::worker { 'worker-hosts-queue':
-      queues => ['hosts_queue'],
+    Integer[1, $hosts_queue_workers].each |$n| {
+      foreman::dynflow::worker { "worker-hosts-queue-${n}":
+        queues => ['hosts_queue'],
+      }
     }
   }
 
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -26,6 +26,8 @@
 #   The contextual package name for the PostgreSQL debversion extension
 # @param postgresql_evr_package
 #   The contextual package name for the PostgreSQL EVR extension
+# @param hosts_queue_workers
+#   The number of dynflow workers for the Katello hosts_queue
 class katello::params (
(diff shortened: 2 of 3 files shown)


ehelms:
Could you open a Redmine issue against the installer for this? @ekohl can talk to the looping.

I find specifying the "ids" to be the odd part vs. specifying a raw number of workers one would like to have.

ehelms:
@ianballou @adamruzicka Do we know where the line between more workers and changing the concurrency is for scaling Dynflow/Sidekiq work for our use cases?

ekohl:
I've seen @adamruzicka asking the performance team to look into this. I'd guess it's also queue dependent (or rather job dependent). If it's a lot of code that's waiting (HTTP requests for example), many threads will scale likely scale better than if there's a lot of actual number crunching. In Puma you can tune the number of workers and the number of threads per worker. With my PR we have the same two options for sidekiq.

adamruzicka:
@ekohl beat me to it. Until we hear back from perf team, I would say the breaking point would be around 10 threads, but I don't have any data to back this claim up.

> This PR will require a change to foreman-maintain to restart all of the running worker-hosts-queue workers.

foreman-maintain should be smart enough to detect the new instances without you having to patch it, but it won't hurt if you check it :)

ianballou:
I'm going to hold of on changes until https://github.com/theforeman/puppet-foreman/pull/843 is in. When it is, I'll just change the `foreman::dynflow::worker` to the new `foreman::dynflow::pool`.

ehelms:
I think we could get a version of this in if you want since the external parameter wouldn't change just the internals once (https://github.com/theforeman/puppet-foreman/pull/843) is merged.

ianballou:
@ehelms I've updated the code to change the worker array to a simple integer.

ianballou:
I'm not too sure what's going on with the Travis but this should be ready.

ianballou:
> We have an installer bug with pulp2 that works in the following way:
> 
>     1. Say you initially configure 8 workers
> 
>     2. At a later time, you decide to reduce the count, say to 6
> 
>     3. The module accepts the new parameter and defines 6 workers, but there are 2 excess workers that never got removed
> 
> 
> @ianballou would you say a similar behavior is possible here?

I could see this happening but I'm not certain.  @ekohl did you need to handle this scenario for your related PR https://github.com/theforeman/puppet-foreman/pull/843 ?

ekohl:
> I could see this happening but I'm not certain. @ekohl did you need to handle this scenario for your related PR [theforeman/puppet-foreman#843](https://github.com/theforeman/puppet-foreman/pull/843) ?

That is indeed what that PR should handle. Right now there are legitimate test failures that I need to resolve, but I want to finish that and rebase this PR to use the new method.

ianballou:
Closing in favor of https://github.com/theforeman/puppet-katello/pull/364

PR #347, opened 2020-06-04
Fixes #30023: Add Katello config for Pulp 3 RPM

ehelms:


jlsherrill:
This looks good to me.  What is it that handles setting this to 'true'  by default on upgrades?

wbclark:
I opened https://github.com/theforeman/foreman-installer/pull/516 but I suspect there may be an issue when a prior commit of that migration has already been run.

PR #351, opened 2020-06-17
Drop certs and ensure ownership of keys

ehelms:
Draft changes from playing around with new proof of concept certificate handling tool. The hardcoded defaults will need updates and exposure of these certificates to be configure directly. This falls in line with the proposed changes to refactor foreman-installer entry points:

https://github.com/theforeman/foreman-installer/pull/421


manifests/application.pp
```diff
@@ -21,31 +21,38 @@
...
+    ignore_missing => false,
+    require        => Foreman::Rake['db:seed'],
+  } ~>
   foreman_config_entry { 'pulp_client_key':
-    value          => $certs::pulp_client::client_key,
+    value          => $pulp_client_key,
     ignore_missing => false,
-    require        => [Class['certs::pulp_client'], Foreman::Rake['db:seed']],
+    require        => Foreman::Rake['db:seed'],
   }
```
ekohl, line 54: Just a thought: can we configure Pulp 3 to use the internal CA for client auth just like we do with Foreman? That way we can use the same Foreman setting to connect to Smart Proxies and simplifies the application overall.

Perhaps @jlsherrill can share his thoughts as well.

manifests/application.pp
```diff
@@ -21,31 +21,38 @@
...
+  include foreman::plugin::tasks
 
-  foreman_config_entry { 'pulp_client_cert':
-    value          => $certs::pulp_client::client_cert,
-    ignore_missing => false,
-    require        => [Class['certs::pulp_client'], Foreman::Rake['db:seed']],
+  file { $candlepin_events_ssl_key:
+    group => $foreman::group,
+    mode  => '0640',
   }
```
ekohl, line 39: Can we also reuse the Foreman client cert, at least by default? Perhaps it needs application changes, but it makes the application easier.

manifests/application.pp
ehelms: The Pulp client certificates use a CN of `pulp` to identify the user who is authenticating. There might be room to switch this to whatever the full DN of the client certificates is. We have client certificates needed for Candlepin Artemis connection as well, and this follows the DN style: https://github.com/theforeman/puppet-certs/blob/master/manifests/candlepin.pp#L32

In the newer certificate setup, I have simplified a lot of this down to really just setting CN since the rest of the attributes are largely arbitrary crap we were putting in there.

So I *think* this could work, if Pulp/Django will allow a "hostname" for a user.

ekohl: Do we only allow a single user (`pulp`) or are there more? If there's just 1, we can let Apache validate and statically pass it to the backend.

ehelms: We only create one user in our implementation of it. I think what you describe is what we do today via:

https://github.com/theforeman/puppet-pulpcore/blob/master/templates/settings.py.erb#L28
https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/templates/pulpcore-api-apache.conf.erb#L4-L7

PR #355, opened 2020-07-08
Fixes #30346: Pass truststore to Candlepin

ehelms:
Requires: https://github.com/theforeman/puppet-certs/pull/291


manifests/candlepin.pp
```diff
@@ -43,7 +43,7 @@
     ca_cert                      => $certs::candlepin::ca_cert,
     keystore_file                => $certs::candlepin::keystore,
     keystore_password            => $certs::candlepin::keystore_password,
-    truststore_password          => $certs::candlepin::keystore_password,
```
ekohl, line 46: The others suggest we need to continue passing this in.

ehelms:
Updated to Fixes and required PRs are merged

ekohl:
Just one thing: do we need to increment the minimum version of puppet-certs for this?

ehelms:
Updated minimum version

metadata.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "name": "katello/certs",
-      "version_requirement": ">= 6.1.1 < 9.0.0"
+      "version_requirement": ">= 8.0.0 < 9.0.0"
```
ekohl, line 29: Technically this should be the next version since it's only in git master. That makes me think we should increase the version in puppet-certs master already, even if we don't release. Looking at the changes there, https://github.com/theforeman/puppet-certs/pull/289#issuecomment-656667661 makes me think we should already bump to 9.0.0. Doing that now also has the benefit that we don't have to do it right before the release.

ehelms: Remind me that process? Is it the same as the release process or is it just literally bumping the version in metadata.json?

ekohl: Right now I'm suggesting to only bump the version in metadata.json. Then we also need to update the consuming modules (katello and foreman_proxy_content) to allow it.

Thinking more about it, I suspect we'll need more breaking changes before we do a real release so let's just go with 9.0.0.

ehelms: Bump: https://github.com/theforeman/puppet-certs/pull/292

ehelms: Bump for FPC -- https://github.com/theforeman/puppet-foreman_proxy_content/pull/273

PR #356, opened 2020-07-09
Fixes #30360: Drop Pulp 2 and Qpid

ehelms:


ehelms:
Related to https://github.com/theforeman/puppet-foreman_proxy_content/pull/271 but does not have the same level of re-factoring impact since all Pulpcore configuration exists over in puppet-foreman_proxy_content.

@jlsherrill can you look and make sure the `katello.yml` is accurate?

manifests/application.pp
```diff
@@ -20,7 +20,6 @@
   Boolean $use_pulp_2_for_file = false,
   Boolean $use_pulp_2_for_docker = false,
   Boolean $use_pulp_2_for_yum = false,
```
ekohl, line 22: These can be dropped as well if there's no Pulp 2.

manifests/application.pp
```diff
@@ -20,7 +20,6 @@
...
-  Stdlib::Absolutepath $repo_export_dir = '/var/lib/pulp/katello-export',
```
ekohl, line 23: Please also drop the docs for this one

templates/katello.yaml.erb
```diff
@@ -7,10 +7,7 @@
   :content_types:
     :file: <%= @enable_file %>
     :yum: <%= @enable_yum %>
-    :deb: <%= @enable_deb %>
-    :puppet: <%= @enable_puppet %>
-    :docker: <%= @enable_docker %>
-    :ostree: <%= @enable_ostree %>
+    :docker: <%= @enable_container %>
```
ekohl, line 10: AFAIK content types are Pulp 2 only. With Pulp 3 they're discovered via capabilities in the Smart Proxy

templates/katello.yaml.erb
```diff
@@ -24,7 +21,6 @@
     :ssl_ca_file: <%= @candlepin_ca_cert %>
 
   :pulp:
-    :url: <%= @pulp_url %>
     :ca_cert_file: <%= @pulp_ca_cert %>
 
   :use_pulp_2_for_content_type:
```
ekohl, line 26: Feels like this could be removed as well

manifests/application.pp
ehelms: I agree they ought to be dropped, just waiting for confirmation from @jlsherrill that they have been dropped code-wise inside Katello.

ekohl: The code suggests that if it's not present, it equals to false so should be safe to drop:
https://github.com/Katello/katello/blob/cf1bf09002252669c997b6317868582d2c35b465/app/models/katello/concerns/smart_proxy_extensions.rb#L149-L156

jlsherrill: yep +1 safe to drop

templates/katello.yaml.erb
jlsherrill: these are technically still needed as this tells katello what nav items to show, and what type of repositories can be created.  We could look into looking at the smart proxy for this, but aren't currently.

templates/katello.yaml.erb
jlsherrill: :+1:

templates/katello.yaml.erb
```diff
@@ -33,8 +29,3 @@
     :yum: <%= @use_pulp_2_for_yum %>
 
   :katello_applicability: <%= !@use_pulp_2_for_yum %>
```
jlsherrill, line 31: this could just be set to 'true' for now

templates/katello.yaml.erb
ekohl: Can we open an issue for that to track? From the installer side we want as little configuration as possible and a more dynamic application would be the result.

PR #357, opened 2020-07-09
Fixes 30362

ehelms:
@ekohl I opened this to try to better capture my thinking as it reflects future certificate handling. I have a sense that this breaks with your original design but this does increase the flexibility of the other manifests. Consider this more of a design in code to get your thoughts on how to enable this.


ekohl:
I think this pretty much aligns on where I've been thinking of going. The only thing is that in the ideal case we would only have a single client cert for Foreman that can be used for Smart Proxy, Pulp and Candlepin. Since it requires a larger effort, I'm fine with that as a long term goal.

If we're talking long term anyway, I think we should also push for `$rest_client_timeout` to be removed. At least from the installer. IMHO it's something that's not required before the application starts and the default should be ok for seeding.

ekohl:
Includes https://github.com/theforeman/puppet-katello/pull/356

PR #358, opened 2020-07-22
Refs #30346 - override candlepin client keypair group

jturel:


jturel:
The related puppet-certs change broke the CandlepinEventListener by changing the client keypair group to tomcat. This overrides it to the foreman user, and it can be overridden to vagrant in dev

jturel:
See also https://github.com/theforeman/puppet-certs/pull/295

jturel:
@ekohl @ehelms thoughts?

ekohl:
I still want to split the class into two. It still deploys files it simply doesn't need at all on the Candlepin server. For those who do deploy Candlepin on a different physical server than Katello, this will break things. We do have those in the community and IMHO should be the default for a dev server - reduces the size of the dev box itself, making rebuilds easier.

jturel:
@ewoud It makes perfect sense, but my goal here is to get nightlies green

jturel:
@ekohl @ehelms can this collection of PRs be merged to address nightly issues?

ehelms:
The failing acceptance test implies you'll need to create the group in this file: https://github.com/theforeman/puppet-katello/blob/master/spec/setup_acceptance_node.pp

ehelms:
I think we need to fix that based on our ability to, and fix what is broken today to unblock the predominant set of users.

ehelms:
Issue to track -- https://projects.theforeman.org/issues/30497

jturel:
> The failing test indicates this is a bad idea and you're breaking the contract that you can deploy Candlepin on a standalone server. The classes should be split into one that is for the server and one that is for the client.

I perceived your prior feedback as an out-of-scope RFE. Nevertheless, I'm struggling to see how it plays out in reality:

- I spin up my Candlepin server with puppet-candlepin, and it has (new) certs::candlepin::server applied
- I spin up my Katello server with puppet-katello and it has (new) certs::candlepin::client applied. The Candlepin server needs to generate the client certs - how is that managed?

Thanks in advance for your detailed explanation

ekohl:
So certs are certainly a weak spot here. You do need to generate a tarball with them beforehand. Sadly puppet-certs is not yet very good at that. It is how @timogoebel and his colleagues at DM have done it in production and they use these classes individually.

jturel:
@ehelms thanks for filing a follow-up issue! the failure in travis now appears unrelated (all but one jobs pass).

@ekohl can you / your team look at fixing the larger issue filed by Eric? I don't feel this change has gone far enough in the wrong direction to not merge it if https://projects.theforeman.org/issues/30497 will be addressed soon. I'd like to see nightly passing this week.

jturel:
puppet6 + el8 failing with:

```
  Generating/signing web server's SSL certificate: localhost-tomcat.crt

  Error: /Stage[main]/Certs::Candlepin/Cert[localhost-tomcat]/ensure: change from 'absent' to 'present' failed: Execution of '/usr/bin/katello-ssl-tool --gen-server --set-hostname localhost --server-cert localhost-tomcat.crt --server-cert-req localhost-tomcat.crt.req --server-key localhost-tomcat.key --server-rpm localhost-tomcat -p file:/etc/pki/katello/private/katello-default-ca.pwd --set-hostname localhost --set-common-name localhost --ca-cert /etc/pki/katello-certs-tools/certs/katello-default-ca.crt --ca-key /etc/pki/katello-certs-tools/private/katello-default-ca.key --set-country US --set-state North Carolina --set-city Raleigh --set-org Katello --set-org-unit SomeOrgUnit --set-email  --cert-expiration 7300' returned 22: ERROR: web server's SSL certificate generation/signing failed:

  

  Using configuration from /root/ssl-build/katello-ca-openssl.cnf

  unable to load number from /root/ssl-build/serial

  error while loading serial number

  139967517005632:error:0D066091:asn1 encoding routines:a2i_ASN1_INTEGER:odd number of chars:crypto/asn1/f_int.c:103:

  

  

  Generating the web server's SSL private key: ./ssl-build/localhost/localhost-tomcat.key

  Rotated: katello-server-openssl.cnf --> katello-server-openssl.cnf.1
```

ehelms:
@jturel That happens sometimes, not sure why, its a bit transient. Now things are green.

PR #359, opened 2020-08-07
Release 15.0.0

wbclark:


wbclark:
I labeled #355 as breaking and #358 as a bugix but I'm not positive in either case.

I allowed newer puppet-candlepin in metadata.json but I'm not sure whether I should increment the required minimum version as well - I think perhaps yes?

ekohl:
> I allowed newer puppet-candlepin in metadata.json but I'm not sure whether I should increment the required minimum version as well - I think perhaps yes?

Yes, the minimum version should also be incremented. In hindsight, we should have bumped candlepin to 10.0.0 in master and https://github.com/theforeman/puppet-katello/pull/355 should have increased the minimum version. This is because it relies on the parameters introduced in 10.0.0. Mind updating? It should be a separate commit with `Refs #30346` in the commit message.

ekohl:
The order needs to be the other way, otherwise git gets confused what's in a release. You can use git rebase -i and force push.

PR #362, opened 2020-09-11
never deploy /pub, that's handled by f_p_c now

evgeni:


ehelms:
This looks mostly like a clean-up and re-factor of which I am OK. The ultimate goal should be for `puppet-foreman_proxy_content` to be the one and only that deploys the pub dir. The problem, which I think was pointed out elsewhere, is the difference in vhost's and the ports we deploy on to. (And who is involved in configuring all of that, Foreman or Pulp)

Do you think we can tackle the centralization of it or does this provide enough value to keep EL8 moving forward?

manifests/pub.pp
```diff
@@ -0,0 +1,15 @@
+# Katello configuration for /pub
+#
+# @param pub_dir_options
+#   The Apache options to use on the `/pub` resource
+#
+class katello::pub(
+  String $pub_dir_options = '+FollowSymLinks +Indexes',
+) {
+  include foreman::config::apache
+  $server_name = $foreman::config::apache::servername
```
ekohl, line 10: This variable is not used in the template so no need to set it.

ekohl (commented):
I have some doubts about this. As @ehelms said, we are planning to work on this and moving twice feels a bit awkward. Especially if long term it's going to disappear completely.

evgeni:
What's the timeline for the centralization? Today, on EL8 there is no /pub so no client can be subscribed as they have no way to obtain the certs, which makes it pretty much an EL8 blocker. I am not 100% sure when we want to have katello (server) support for el8, but probably sooner than later? content proxy is another beast to tackle then.

evgeni:
FWIW, this is not the *only* issue I am facing on EL8 (like e.g. https://projects.theforeman.org/issues/30832) but it's kind of the root of everything :D

ekohl (approved):
:+1: as long as the FPC PR is merged.

PR #363, opened 2020-09-25
don't configure Pulp2 when running on EL8

evgeni:


templates/katello.yaml.erb
```diff
@@ -23,9 +23,11 @@
     :ssl_key_file: <%= @candlepin_events_ssl_key %>
     :ssl_ca_file: <%= @candlepin_ca_cert %>
 
+<% if @pulp2_support %>
   :pulp:
     :url: <%= @pulp_url %>
     :ca_cert_file: <%= @pulp_ca_cert %>
+<% end %>
 
   :use_pulp_2_for_content_type:
```
ehelms, line 32: We probably should drop this section too.

ehelms: And ensure   :katello_applicability: <%= !@use_pulp_2_for_yum %> is true

ehelms: @jlsherrill what happens to https://github.com/theforeman/puppet-katello/pull/363/files#diff-c48970cf4dfa5d2a2124a59cac2bf934R40 in a pure Pulp 3 scenario?

templates/katello.yaml.erb
```diff
@@ -33,6 +34,9 @@
     :yum: <%= @use_pulp_2_for_yum %>
 
   :katello_applicability: <%= !@use_pulp_2_for_yum %>
+<% else %>
+  :katello_applicability: true
+<% end %>
 
   # Internal configuration for communication from server to pulp crane service.
```
evgeni, line 41: crane is pulp2-only too, right?

ehelms: yes

manifests/application.pp
```diff
@@ -66,6 +66,12 @@
   $postgresql_evr_package = $katello::params::postgresql_evr_package
   $manage_db = $foreman::db_manage
 
+  if $facts['os']['release']['major'] == '7' {
+    $pulp2_support = true
+  } else {
+    $pulp2_support = false
+  }
```
ekohl, line 73: I think this also works
```suggestion
  $pulp2_support = $facts['os']['release']['major'] == '7'
```

spec/classes/application_spec.rb
```diff
@@ -60,38 +60,67 @@ class { 'katello::params':
...
-            '    :file: false',
-            '    :yum: false',
-            '  :katello_applicability: true',
-            '  :container_image_registry:',
-            '    :crane_url: https://foo.example.com:5000',
-            '    :crane_ca_cert_file: /etc/pki/katello/certs/katello-server-ca.crt'
-          ])
+        if facts[:operatingsystemmajrelease] == '7'
+          it 'should generate correct katello.yaml' do
+            verify_exact_contents(catalogue, '/etc/foreman/plugins/katello.yaml', [
```
ekohl, line 65: Slightly different way would be:
```ruby
let(:katello_yaml_content) do
    []
  else
    []
  end
end

it 'should generate correct katello.yaml' do
  verify_exact_contents(catalogue, '/etc/foreman/plugins/katello.yaml', katello_yaml_content)
end
```

evgeni:
FWIW, even after dropping all these, the app will have them configured, because it has defaults:

https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L37-L45
https://github.com/Katello/katello/blob/master/lib/katello/engine.rb#L27

manifests/application.pp
```diff
@@ -66,6 +66,8 @@
   $postgresql_evr_package = $katello::params::postgresql_evr_package
   $manage_db = $foreman::db_manage
 
+  $pulp2_support = $facts['os']['release']['major'] == '7'
```
ekohl, line 69: Should this be a parameter? That way in `init.pp` we can also set it. This may be useful in the future to test on pure Pulp 3 setups on EL7.

evgeni: Like this?

ekohl (commented):
Pretty much. If this would also use it here:
https://github.com/theforeman/puppet-katello/blob/45b5dc32a805abbeb5d9629b961703b1071cc004/manifests/init.pp#L187

ekohl (approved):
Let's see what Travis says about the tests. Also a note to the person who merges this: it should be squashed on merge.

PR #364, opened 2020-10-19
Refs #29817 - Implement a resizable hosts_queue pool

ekohl:
This uses the new foreman::dynflow::pool defined type which allows users to easily add additional workers or scale them back.

Currently a draft since it's pointing to my branch in the fixtures to allow testing. Blocked on https://github.com/theforeman/puppet-foreman/pull/843.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -3,8 +3,12 @@
 # @param rest_client_timeout
 #   Timeout for Katello rest API
 #
+# @param hosts_queue_workers
+#   The number of workers handling the hosts_queue queue.
+#
 class katello::application (
   Integer[0] $rest_client_timeout = 3600,
+  Integer[0] $hosts_queue_workers = 1,
 ) {
   include foreman
   include certs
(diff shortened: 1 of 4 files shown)


ehelms:
https://github.com/theforeman/puppet-foreman/pull/843 allows users to control the concurrency and the number of workers, here I only see number of workers, is that an intended design to restrict it for this use case?

ekohl:
Not intentional, just a rushed patch here. @ianballou could you have a look about the desired config options and if this properly replaces your patch?

Also looks like I messed up the fixtures. Will update both once there's clarity about which params are expected.

ianballou:
@ekohl, it looks like this PR sufficiently covers my old one :+1:

ianballou:
I'm curious what's going on with this PR, I haven't checked-in for a while.  As far as I know we'd still like this functionality for Pulp 3, but someone can let me know if we don't.

ehelms:
This can now be updated with https://github.com/theforeman/puppet-foreman/pull/843 having been merged

ekohl:
Rebased to resolve merge conflicts and it depends on theforeman/foreman 17.x to ensure the correct API is present.

ehelms (approved):

PR #365, opened 2020-10-22
Do not enable Puppet if there is no Pulp 2

ehelms:


manifests/params.pp
```diff
@@ -39,4 +39,10 @@
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
   Boolean $pulp2_support = $katello::globals::pulp2_support,
 ) inherits katello::globals {
+
+  if !$pulp2_support {
+    $_enable_puppet = false
+  } else {
+    $_enable_puppet = $enable_puppet
+  }
```
ekohl, line 47:
```suggestion
  $_enable_puppet = $pulp2_support and $enable_puppet
```

manifests/application.pp
```diff
@@ -50,7 +50,7 @@
   $enable_ostree = $katello::params::enable_ostree
   $enable_yum = $katello::params::enable_yum
   $enable_file = $katello::params::enable_file
-  $enable_puppet = $katello::params::enable_puppet
+  $enable_puppet = $katello::params::$_enable_puppet
```
ekohl, line 53:
```suggestion
  $enable_puppet = $katello::params::_enable_puppet
```

manifests/params.pp
```diff
@@ -39,4 +39,10 @@
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
   Boolean $pulp2_support = $katello::globals::pulp2_support,
 ) inherits katello::globals {
+
```
ekohl, line 42: This empty line breaks CI once we update to the latest lint module. Please drop it.

manifests/params.pp
```diff
@@ -39,4 +39,10 @@
...
+  if !$pulp2_support {
```
ekohl, line 43: I avoid the pattern `if !$bool { first } else { second }` and always write `if $bool { second } else { first }`. However, in this case the whole if can be avoided. See the suggestion below.

manifests/params.pp
ehelms: That still makes me sad :/ The empty line provides a nice break when reading the code.

manifests/pulp.pp
```diff
@@ -146,7 +146,7 @@
     enable_rpm             => $katello::params::enable_yum,
     enable_iso             => $katello::params::enable_file,
     enable_deb             => $katello::params::enable_deb,
-    enable_puppet          => $katello::params::enable_puppet,
+    enable_puppet          => $katello::params::_enable_puppet,
```
ekohl, line 149: This part doesn't make a lot of sense. By definition we have Pulp 2 support at this point.
```suggestion
    enable_puppet          => $katello::params::enable_puppet,
```

manifests/application.pp
```diff
@@ -50,7 +50,7 @@
   $enable_ostree = $katello::params::enable_ostree
   $enable_yum = $katello::params::enable_yum
   $enable_file = $katello::params::enable_file
-  $enable_puppet = $katello::params::enable_puppet
+  $enable_puppet = $katello::params::_enable_puppet
```
ekohl, line 53: Thinking more about it, should we fix it here:
```suggestion
  $enable_puppet = $katello::params::pulp2_support and $katello::params::enable_puppet
```

That avoids the additional variable.

manifests/params.pp
```diff
@@ -39,4 +39,5 @@
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
   Boolean $pulp2_support = $katello::globals::pulp2_support,
 ) inherits katello::globals {
+  $_enable_puppet = $pulp2_support and $enable_puppet
```
ekohl, line 42: If both other suggestions are accepted, this can be dropped
```suggestion
```

ekohl (commented):
Apologies for not spotting this in the earlier review.

manifests/application.pp
ehelms: Sure, I really had no idea, this module becoming really hard to follow after the re-factor to params/globals/application. Hopefully we can just drop all of that organization looking ahead.

ekohl: The long term goal I had was to stop using init.pp altogether. Then `globals` would be exposed in kafo as well as `application` and `candlepin`. `params` wouldn't be. https://github.com/theforeman/foreman-installer/pull/421 would do that. I rebased it, but it needs migrations. However, what that allows us to do is to remove the `pulp` module from the answers file on EL8, which means you only have `--pulp-*` options on EL7. Due to different priorities I was never able to finish this work.

ekohl (approved):

PR #367, opened 2020-10-29
Switch to GH Actions from Travis

ehelms:
Travis has slowed to a crawl this week and is impacting releases. This is based on current (as of this posting) state of https://github.com/theforeman/foreman-installer-modulesync/pull/106

I figure, there may be iterations to go on the modulesync PR, but if this passes we can do those updates after the fact and unblock release activities.


wbclark (commented):
This enables GH Actions but doesn't disable Travis. So currently we are waiting on Travis :)

Is it intentional to get some overlap for now to see everything is green and would you then add a 2nd commit to this that also disables Travis?

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,2 @@
+# This module contains no acceptance tests - this is a placeholder for
+# modulesync.
```
wbclark, line 2: With this being an empty placeholder for now, disabling Travis means we would have no acceptance tests in CI.

So, the plan would be that we are running these locally only during the meantime?

ehelms: good catch, this was a bad sync on my part and thus it got rendered incorrectly

ehelms: Fixed now

ehelms:
I think I would just leave Travis for now, and ignore it's results just to do it in two steps?

wbclark:
> I think I would just leave Travis for now, and ignore it's results just to do it in two steps?

We have reduced concurrency with Travis at the moment due to https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions

So if we have any other repos where we are still relying on Travis and not yet migrating to GH Actions, then we are creating additional unnecessary load on Travis here which will exacerbate the problem for our other projects.

Therefore my thinking is we should disable it unless we will make this same change on every repository across our Travis CI account.

ehelms:
This has dropped the Travis file now, we can always put it back as you say

wbclark (approved):
Thanks @ehelms !

PR #370, opened 2020-11-03
Support use_pulp_2_for_deb

jlsherrill:


jlsherrill:
we still need to handle apache configuration, and upgrade/migrations but this is a start

jlsherrill:
@sbernhard  pinging for awareness

sbernhard:
Thank you for letting me know.

ekohl:
> we still need to handle apache configuration, and upgrade/migrations but this is a start

FYI: we greatly refactored how we do this. It's now all in [puppet-pulpcore](https://github.com/theforeman/puppet-pulpcore) and plugins provide a single boolean (like [rpm](https://github.com/theforeman/puppet-pulpcore/blob/bb0b65a69dbada288b069494e45a1800745c5f7d/manifests/plugin/rpm.pp#L5)). That's set in [puppet-foreman_proxy_content](https://github.com/theforeman/puppet-foreman_proxy_content) (like [here](https://github.com/theforeman/puppet-foreman_proxy_content/blob/a620f4115514a9fed26c4815e9eff1442b380c87/manifests/init.pp#L416-L418)).

ekohl (approved):
I'm not sure if this is still blocked on anything.

sbernhard (approved):

jlsherrill:
I think all we need still is the upgrade/migration in foreman-installer, but its probably not blocked on it.  I'll try to get to that today

PR #371, opened 2020-11-24
Drop Crane configuration that is no longer used from Pulp 2

ehelms:


templates/katello.yaml.erb
```diff
@@ -35,10 +35,6 @@
 
   :katello_applicability: <%= !@use_pulp_2_for_yum %>
 
-  # Internal configuration for communication from server to pulp crane service.
-  :container_image_registry:
-    :crane_url: <%= @crane_url %>
-    :crane_ca_cert_file: <%= @crane_ca_cert %>
```
ekohl, line 41: I recently found out that this never worked since the code was looking for a different one.

ekohl (approved):

PR #372, opened 2021-01-08
Refs #31346 - Use Java 11 for Candlepin

jturel:


jturel:
@ekohl @ehelms based on the latest discussion in https://github.com/theforeman/puppet-candlepin/pull/169 is this the correct interpretation / should we do this vs the puppet-candlepin change?

ekohl (changes requested):
You are missing the dependency chaining. There's 2 ways we can do this. Either here in this class or introduce a parameter to candlepin for the java package (`Optional[String[1]] $java_package = undef`) and add that to `candlepin::install` in the right order. I'm slightly leaning to the latter but wouldn't mind the former either. As long as we ensure that the java package is installer before installing candlepin.

jturel:
[test]

jturel:
@ekohl good to merge? :)

ekohl:
I would like to get a release of puppet-candlepin our and update metadata.json in this module to require at least that since it's a new parameter.

jturel:
Alright, let me know what -candlepin version we end up on and I can update metadata.json here

ekohl:
That would be https://github.com/theforeman/puppet-candlepin/pull/171

ekohl (changes requested):
Small nit: I think we should change the commit message (and PR title) to state that Candlepin will be using Java 11. Other than that, :+1:

jturel:
@ekohl updated!

ekohl (approved):
Isn't this `Fixes #31346` instead of `Refs`?

jturel:
> Isn't this `Fixes #31346` instead of `Refs`?

The redmine has a collection of PRs to deliver the desired outcome, so that's why I used refs (this PR alone does not get us to where we need to be).

ekohl:
It feels like this is the one that really makes it happen, but I'm fine either way. Let me know if this should be merged as is or you want to change it.

jturel:
> It feels like this is the one that really makes it happen, but I'm fine either way. Let me know if this should be merged as is or you want to change it.

I see your point, but I like how Refs ommunicates that there's more to the story. Let's merge as-is

PR #373, opened 2021-01-12
Drop enabling Pulp 2 content types

ehelms:
At this point, on the server, Pulp 3 handles all content. I am curious to see if these being enabled and putting their Apache configs in place are interfering with content syncs. I would like to try and eliminate this as the cause of: https://ci.centos.org/job/foreman-pipeline-katello-nightly-centos7-upgrade/215/tapTestReport/debug%2Fn28.pufty%2Ftmp%2Fdebug-katello-nightly-centos7%2Fpipe-up-katello-nightly-centos7%2Froot%2Fbats_results%2Ffb-katello-proxy.bats.tap-3/


ekohl (approved):
Should this reference the Redmine #31614 as well?

ekohl (changes requested):
Oh, this should then also drop `$use_pulp_2_for_file` and friends as parameters since they for sure can't be used anymore.

ehelms:
Dropped the `use_pulp` parameters

templates/katello.yaml.erb
```diff
@@ -28,12 +28,6 @@
     :url: <%= @pulp_url %>
     :ca_cert_file: <%= @pulp_ca_cert %>
 
-  :use_pulp_2_for_content_type:
-    :docker: <%= @use_pulp_2_for_docker %>
-    :file: <%= @use_pulp_2_for_file %>
-    :yum: <%= @use_pulp_2_for_yum %>
-    :deb: <%= @use_pulp_2_for_deb %>
-
   :katello_applicability: <%= !@use_pulp_2_for_yum %>
```
ekohl, line 31: `@use_pulp_2_for_yum` is now undefined
```suggestion
  :katello_applicability: true
```

Since it'll be true in both branches of `if @pulp2_support`, it can be moved out of it.

templates/katello.yaml.erb
```diff
@@ -23,19 +23,10 @@
     :ssl_key_file: <%= @candlepin_events_ssl_key %>
     :ssl_ca_file: <%= @candlepin_ca_cert %>
 
+  :katello_applicability: true
```
ekohl, line 26: @jlsherrill does it make sense for Katello to become smarter about this? Looks like today it must be set but perhaps this should default to true?

templates/katello.yaml.erb
```diff
@@ -23,19 +23,10 @@
...
+
 <% if @pulp2_support %>
```
ekohl, line 28: Small nit: you get cleaner output if you swap it around:
```suggestion
<% if @pulp2_support %>

```

The net result is that for users where `@pulp2_support` is false, is that the file doesn't change for them. That means it may not have to restart Foreman(/Katello).

ekohl (approved):

templates/katello.yaml.erb
ehelms: I believe eventually there will be more clean up in this file as Pulp 2 is completely removed.

templates/katello.yaml.erb
ehelms: I do not think we have any users for which that is true right now, and this file will undergo further changes in the end to support Katello 4.0, so I will hold off further changes.

PR #376, opened 2021-01-15
Drop inclusion of pulp-vhosts80 with no managed content types

ehelms:
Follow on for: a0917838fabc8241e3e5e7f678378a2d45cbea54

### DIFF
--- a/templates/pulp-apache.conf.erb
+++ b/templates/pulp-apache.conf.erb
@@ -1,7 +1,5 @@
 ### File managed with puppet ###
 
-Include <%= scope['apache::confd_dir'] %>/pulp-vhosts80/*.conf
-
 <Location /pulp>
   <IfModule mod_passenger.c>
     PassengerEnabled off


evgeni (approved):

templates/pulp-apache.conf.erb
```diff
@@ -1,7 +1,5 @@
 ### File managed with puppet ###
 
-Include <%= scope['apache::confd_dir'] %>/pulp-vhosts80/*.conf
-
 <Location /pulp>
```
ekohl, line 3: If Pulp is no longer serving content but it's all done by Pulpcore, is this part still useful or should the entire file be dropped?

ehelms: I left this part, as Pulp is still handling katello-agent which does require API requests to Pulp presently. This should get cleaned up along with fully dropping Pulp once the switchover has been made.

PR #377, opened 2021-01-19
Fixes #31693 - Restore katello::qpid_client

jturel:
This reverts commit 990b7d2c78a31d271064e912760f8dc0fe0a6f2d.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -18,6 +18,12 @@ class katello::application (
   include certs::pulp_client
   include katello::params
 
+  include certs::qpid
+  include katello::qpid_client
+  User<|title == $foreman::user|>{groups +> 'qpidd'}
+
+  Class['certs', 'certs::ca', 'certs::qpid'] ~> Class['foreman::plugin::tasks']
+
   foreman_config_entry { 'pulp_client_cert':
     value          => $certs::pulp_client::client_cert,
     ignore_missing => false,
--- a/spec/classes/application_spec.rb
+++ b/spec/classes/application_spec.rb
@@ -28,6 +28,13 @@ describe 'katello::application' do
         end
 
         it do
+          is_expected.to contain_class('certs::qpid')
+            .that_notifies(['Class[Foreman::Plugin::Tasks]'])
+        end
+
+        it { is_expected.to contain_class('katello::qpid_client') }
+
+        it do
           is_expected.to create_foreman_config_entry('pulp_client_cert')
             .with_value('/etc/pki/katello/certs/pulp-client.crt')
             .that_requires(['Class[Certs::Pulp_client]', 'Foreman::Rake[db:seed]'])
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -8,6 +8,7 @@ describe 'katello' do
       it { is_expected.to compile.with_all_deps }
       it { is_expected.to contain_class('katello::candlepin') }
       it { is_expected.to contain_class('katello::application') }
+      it { is_expected.to contain_class('katello::qpid_client') }
 
       if facts[:operatingsystemmajrelease] == '7'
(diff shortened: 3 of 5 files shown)


jturel:
This works for me in testing, seeing if the build passes

jturel:
@ekohl this commit is a revert but maybe it should have its own issue. What do you think?

ehelms:
I wouldn't mind an installer issue linked to the Katello issue from https://github.com/Katello/katello/pull/9088 to track both.

jturel:
> I wouldn't mind an installer issue linked to the Katello issue from [Katello/katello#9088](https://github.com/Katello/katello/pull/9088) to track both.

Done!

spec/classes/qpid_client_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe 'katello::qpid_client' do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let(:facts) { facts }
+
+      if facts[:operatingsystemmajrelease] == '7'
```
ekohl, line 8: A better practice is to not shadow variables. You can also add conditionals to a context.
```suggestion
  on_supported_os.each do |os, facts|
    context "on #{os}", if: os_facts[:operatingsystemmajrelease] == '7' do
      let(:facts) { os_facts }

```

manifests/application.pp
```diff
@@ -18,6 +18,14 @@
   include certs::pulp_client
   include katello::params
 
+  if $facts['os']['release']['major'] == '7' {
```
ehelms, line 21: We can drop this condition now right @jturel ?

spec/classes/qpid_client_spec.rb
jturel: This check is no longer there!

jturel:
@ekohl @ehelms thanks for the review - updated!

ehelms (approved):
LGTM

spec/classes/init_spec.rb
```diff
@@ -11,6 +11,7 @@
 
       if facts[:operatingsystemmajrelease] == '7'
         it { is_expected.to contain_class('katello::pulp') }
+        it { is_expected.to contain_class('katello::qpid_client') }
```
ekohl, line 14: This isn't EL7 specific so in the test it shouldn't be conditional.

ekohl (changes requested):

spec/classes/init_spec.rb
jturel: nice catch

ekohl: This looks swapped. I think `katello::qpid` is EL7-only and `katello::qpid_client` is all versions. At least, https://github.com/theforeman/puppet-katello/blob/a0917838fabc8241e3e5e7f678378a2d45cbea54/manifests/init.pp#L174-L178 implies that.

jturel: Yep :)

spec/classes/application_spec.rb
```diff
@@ -21,12 +21,20 @@ class { 'katello::params':
           it { is_expected.to create_package('tfm-rubygem-katello') }
           it { is_expected.not_to create_package('tfm-rubygem-katello').that_requires('Anchor[katello::candlepin]') }
           it { is_expected.to create_package('rh-postgresql12-postgresql-evr') }
+
```
ekohl, line 24: Nit: is this change needed?

ekohl (approved):
Small nit, otherwise :+1:

spec/classes/application_spec.rb
jturel: Fixed

ehelms:
Thanks @jturel -- merging so I can include in the release

PR #380, opened 2021-02-01
Fixes #30360: Remove Pulp 2 deployment code

ehelms:
Needed as part of the move to Katello sending and receiving katello-agent messages (https://github.com/Katello/katello/pull/9088)


ekohl:
Is this supposed to be part of Katello 4.0?

ehelms:
I believe it will have to be as part of the katello-agent changes.

ekohl (changes requested):
On a related note: I'd like for Katello to determine the content types dynamically based on what's on the Smart Proxy. That would remove a lot of these config options.

I also think that these options are no longer really global if we drop Pulp 2: they're just for katello. That means I'd prefer to move them around.

My reasoning is that globals.pp are for options that are shared between multiple systems (Katello, Pulp, Candlepin). Now that Pulp is being moved to the content proxy module entirely, there are fewer of those options. Ideally we'd have none and every system is independent.

We also have the "system" classes (application, candlepin, qpid) that configure just a system. Any option that's just for one system is a parameter on that class.

init.pp is a wrapper that I would like to get rid of.

Lastly there's params which are things you can tune, but are really advanced.

So https://github.com/theforeman/foreman-installer/pull/421 does this by exposing `katello::globals`, `katello::application`, `katello::candlepin` etc. They are mapped to the top level to drop the `katello-` prefix.

I'd like to have this PR in such a shape that it works towards that goal. For now that means getting the class parameters set in the right places.

ehelms:
> On a related note: I'd like for Katello to determine the content types dynamically based on what's on the Smart Proxy. That would remove a lot of these config options.

They are aware of making this change, it requires more work within the Katello code.



> I'd like to have this PR in such a shape that it works towards that goal. For now that means getting the class parameters set in the right places.

I am OK with some small changes or clean up in this PR, but my aim here is not for this to solve any module restructuring. For Katello to switch to owning katello-agent, we'll want (possibly need) the Pulp 2 bits not present on the system and I think we should have a code change that does only that. Further re-factoring can then happen after the functional change.

ekohl:
> I am OK with some small changes or clean up in this PR, but my aim here is not for this to solve any module restructuring. For Katello to switch to owning katello-agent, we'll want (possibly need) the Pulp 2 bits not present on the system and I think we should have a code change that does only that. Further re-factoring can then happen after the functional change.

My point is that as you changed it now (in particular `enable_puppet`), it won't work as you expect. Hopefully with the design behind it, you can see why it doesn't work. So rather than keeping the dead `enable_puppet` parameter, it should be dropped. You can even kill it completely and hardcode it in the template.

ekohl (approved):

manifests/application.pp
```diff
@@ -3,12 +3,8 @@
 # @param rest_client_timeout
 #   Timeout for Katello rest API
 #
-# @param repo_export_dir
-#   Create a repository export directory for Katello to use
-#
 class katello::application (
   Integer[0] $rest_client_timeout = 3600,
```
ekohl, line 7: Unrelated, but I think this is also a good candidate to move to a regular setting that can be changed at runtime. Do you agree and should I create an issue for it?

ehelms: Let's see what @jlsherrill  thinks

PR #386, opened 2021-02-19
Require puppet-certs 11+

ehelms:


evgeni (approved):

metadata.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "name": "katello/certs",
-      "version_requirement": ">= 9.0.0 < 11.0.0"
+      "version_requirement": ">= 11.0.0 < 12.0.0"
```
ekohl, line 29: Why is version 11 required? Looking at the release log all the changes were in areas we don't touch in this module so normally I would expect it to be `>= 9.0.0 < 12.0.0`.

PR #387, opened 2021-02-19
Refs #32037: Add flag to enable katello_agent infrastructure and disable it by default

ehelms:
…ault

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -41,5 +41,6 @@ class katello::params (
   $enable_file = $katello::globals::enable_file != false
   $enable_docker = $katello::globals::enable_docker != false
   $enable_deb = $katello::globals::enable_deb != false
+  $enable_katello_agent = $katello::globals::enable_katello_agent != false
 
 }
--- a/manifests/qpid.pp
+++ b/manifests/qpid.pp
@@ -10,31 +10,50 @@ class katello::qpid (
   String $interface = 'lo',
   Integer[0, 5000] $wcache_page_size = 4,
 ) {
-  include certs
-  include certs::qpid
   include katello::params
 
   $agent_event_queue_name = $katello::params::agent_event_queue_name
 
+  if $katello::params::enable_katello_agent {
+    include certs
+    include certs::qpid
+
+    $_qpid_ensure = 'present'
+    $ssl = true
+    $ssl_cert_db = $certs::qpid::nss_db_dir
+    $ssl_cert_password_file = $certs::qpid::nss_db_password_file
+    $ssl_cert_name = $certs::qpid::nss_cert_name
+
+    Class['certs', 'certs::qpid'] ~> Class['qpid']
+  } else {
+    $_qpid_ensure = 'absent'
+    $ssl = false
+    $ssl_cert_db = undef
+    $ssl_cert_password_file = undef
+    $ssl_cert_name = undef
+  }
+
   class { 'qpid':
-    ssl                    => true,
-    ssl_cert_db            => $certs::qpid::nss_db_dir,
-    ssl_cert_password_file => $certs::qpid::nss_db_password_file,
-    ssl_cert_name          => $certs::qpid::nss_cert_name,
+    ensure                 => $_qpid_ensure,
+    ssl                    => $ssl,
+    ssl_cert_db            => $ssl_cert_db,
+    ssl_cert_password_file => $ssl_cert_password_file,
+    ssl_cert_name          => $ssl_cert_name,
     acl_content            => template('katello/qpid_acls.acl'),
     interface              => $interface,
     wcache_page_size       => $wcache_page_size,
     auth                   => true,
-    subscribe              => Class['certs', 'certs::qpid'],
   }
 
   contain qpid
 
-  qpid::config::queue { $katello::params::agent_event_queue_name:
-    ssl_cert       => $certs::qpid::client_cert,
-    ssl_key        => $certs::qpid::client_key,
-    hostname       => $katello::params::qpid_hostname,
-    username       => $certs::qpid::hostname,
-    sasl_mechanism => 'EXTERNAL',
+  if $katello::params::enable_katello_agent {
+    qpid::config::queue { $katello::params::agent_event_queue_name:
+      ssl_cert       => $certs::qpid::client_cert,
+      ssl_key        => $certs::qpid::client_key,
+      hostname       => $katello::params::qpid_hostname,
+      username       => $certs::qpid::hostname,
+      sasl_mechanism => 'EXTERNAL',
+    }
   }
 }
--- a/metadata.json
+++ b/metadata.json
@@ -30,7 +30,7 @@
     },
     {
       "name": "katello/qpid",
-      "version_requirement": ">= 7.0.0 < 9.0.0"
+      "version_requirement": ">= 8.0.0 < 9.0.0"
     },
     {
       "name": "theforeman/foreman",
--- a/spec/acceptance/qpid_spec.rb
+++ b/spec/acceptance/qpid_spec.rb
@@ -1,20 +1,59 @@
 require 'spec_helper_acceptance'
 
 describe 'Install qpid' do
-  let(:pp) { 'include katello::qpid' }
+  context 'with enable_katello_agent true' do
+    let(:pp) do
+      <<-PUPPET
+      class { 'katello::globals':
+        enable_katello_agent => true,
+      }
+      include katello::qpid
+      PUPPET
+    end
 
-  it_behaves_like 'a idempotent resource'
+    it_behaves_like 'a idempotent resource'
 
-  describe service('qpidd') do
-    it { is_expected.to be_running }
-    it { is_expected.to be_enabled }
+    describe service('qpidd') do
+      it { is_expected.to be_running }
+      it { is_expected.to be_enabled }
+    end
+
+    describe port('5671') do
+      it { is_expected.to be_listening }
+    end
+
+    describe port('5672') do
+      it { is_expected.to be_listening }
+    end
   end
 
-  describe port('5671') do
-    it { is_expected.to be_listening }
-  end
+  context 'with enable_katello_agent false' do
+    let(:pp) do
+      <<-PUPPET
+      class { 'katello::globals':
+        enable_katello_agent => false,
+      }
+      include katello::qpid
+      PUPPET
+    end
 
-  describe port('5672') do
-    it { is_expected.to be_listening }
+    it_behaves_like 'a idempotent resource'
+
+    describe service('qpidd') do
+      it { is_expected.not_to be_running }
+      it { is_expected.not_to be_enabled }
+    end
+
+    describe port('5671') do
+      it { is_expected.not_to be_listening }
+    end
+
+    describe port('5672') do
+      it { is_expected.not_to be_listening }
+    end
+
+    describe package('qpid-cpp-server') do
+      it { is_expected.not_to be_installed }
+    end
   end
 end
--- a/spec/classes/application_spec.rb
+++ b/spec/classes/application_spec.rb
@@ -81,8 +81,7 @@ describe 'katello::application' do
               '    :ssl_key_file: /etc/pki/katello/private/java-client.key',
               '    :ssl_ca_file: /etc/pki/katello/certs/katello-default-ca.crt',
               '  :agent:',
-              '    :broker_url: amqps://localhost:5671',
-              '    :event_queue_name: katello.agent',
+              '    :enabled: false',
               '  :katello_applicability: true',
             ]
           else
@@ -106,8 +105,7 @@ describe 'katello::application' do
               '    :ssl_key_file: /etc/pki/katello/private/java-client.key',
               '    :ssl_ca_file: /etc/pki/katello/certs/katello-default-ca.crt',
               '  :agent:',
-              '    :broker_url: amqps://localhost:5671',
-              '    :event_queue_name: katello.agent',
+              '    :enabled: false',
               '  :katello_applicability: true',
             ]
           end
@@ -161,8 +159,7 @@ describe 'katello::application' do
               '    :ssl_key_file: /etc/pki/katello/private/java-client.key',
               '    :ssl_ca_file: /etc/pki/katello/certs/katello-default-ca.crt',
               '  :agent:',
-              '    :broker_url: amqps://localhost:5671',
-              '    :event_queue_name: katello.agent',
+              '    :enabled: false',
               '  :katello_applicability: true',
             ]
           else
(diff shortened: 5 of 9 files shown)


ehelms:
Initial, simplest version of having a way to turn off katello-agent qpid and qpid-dispatch infrastructure and to disable it as the default experience. This would be used in conjunction with https://github.com/theforeman/puppet-foreman_proxy_content/pull/329.

For installations use the installer, this would result in users turning it on via:

Server:
```
foreman-installer --katello-enable-katello-agent true --foreman-proxy-content-enable-katello-agent true
```

Proxy:
```
foreman-installer --foreman-proxy-content-enable-katello-agent true
```

As this introduces a new parameter, we will need a migration to ensure upgrades do not turn this infrastructure off by default. And in the upgrade guide we can mention that if users do not use katello-agent they can turn this off now and save themselves some resources. This will need support inside Katello as well that @jturel is working on.

The biggest downside to this implementation is that it requires two parameters to disable or enable on a server, but it is low cost in the amount of changes required.

manifests/init.pp
```diff
@@ -4,13 +4,15 @@
...
-# $enable_file::        Enable generic file content management
+# $enable_file::          Enable generic file content management
 #
-# $enable_docker::      Enable docker content plugin
+# $enable_docker::        Enable docker content plugin
 #
-# $enable_deb::         Enable debian content plugin
+# $enable_deb::           Enable debian content plugin
+#
+# $enable_katello_agent:: Enable use of katello-agent for remote yum actions
```
ekohl, line 15: I wonder if it makes sense to name it
```suggestion
# $enable_agent:: Enable use of katello-agent for remote yum actions
```

ekohl (commented):
Overall this looks like it makes sense.

manifests/init.pp
ehelms: I know it might seem redundant, since the module has katello already in and it the parameter translated in the installer would also start with katello, but I feel like katello-agent is so widely known as a term that we shouldn't shorten it to just agent. I would want no ambiguity here.

ehelms:
Now requires https://github.com/theforeman/puppet-qpid/pull/148

manifests/init.pp
wbclark: This reasoning is sound to me. Particularly when an user can encounter these parameters in a setting (foreman-installer) which is pretty removed from these modules, I see a clear benefit to being a little verbose with parameter names.

manifests/qpid.pp
```diff
@@ -6,9 +6,17 @@
...
+#   Specify if qpidd service should be running or stopped
+#
+# @param service_enable
+#   Enable qpidd service at boot
+#
 class katello::qpid (
   String $interface = 'lo',
   Integer[0, 5000] $wcache_page_size = 4,
+  Boolean service_enable = false,
+  Boolean service_ensure = false,
```
ekohl, line 19: What's the use case for splitting these at this level?

Also, I'd expect this to reuse the variable via params as a default. That way you can set it via globals to set the behavior.

ekohl (changes requested):

manifests/qpid.pp
ehelms: > What's the use case for splitting these at this level?

Mostly I'm just trying my best to follow patterns to ensure maximum flexibility as that always has seemed to be our modus operandi. I don't have an opinion which way I go with how many parameters or where they are exposed at. 

> Also, I'd expect this to reuse the variable via params as a default. That way you can set it via globals to set the behavior.

I still don't understand the params and global pattern in this module, so you'll need to explain to it to me like I am five (or just throw code example applied here at me).

manifests/init.pp
```diff
@@ -102,5 +107,7 @@
   class { 'katello::qpid':
     interface        => $qpid_interface,
     wcache_page_size => $qpid_wcache_page_size,
+    serivce_enable   => $enable_katello_agent,
```
ekohl, line 110:
```suggestion
    service_enable   => $enable_katello_agent,
```

ehelms:
A question for testing in our pipelines with this change is how we continue to test katello-agent related client actions. The options as I see them:

 * Update Forklift pipelines to set the enable katello-agent flags on the installer at install time, bats run as is
 * Re-run the installer with enable katello-agent flags in bats and then run the katello-agent client tests following that

ehelms:
Needs https://github.com/theforeman/puppet-katello/pull/395

wbclark:
This needs a puppet-qpid release ( https://github.com/theforeman/puppet-qpid/pull/151 ) to properly test it

Once that happens it also needs the dependency bumped to get the version which added that feature

wbclark:
> * Update Forklift pipelines to set the enable katello-agent flags on the installer at install time, bats run as is
> * Re-run the installer with enable katello-agent flags in bats and then run the katello-agent client tests following that

I somewhat like the second option better as we should test the default installation and this seems more aligned with the longer term goal of replacing katello-agent. With the approach of later re-running the installer to enable the feature and then run katello-agent client tests, there is a gap in that we don't really ensure this feature doesn't break other tests, but I suppose it's never going to be feasible to test every possible combination of installer answers.

evgeni:
"running installer in bats" sounds awful, but not testing the default scenario is worse.

if only we had a way to test different server types, similarly we can for clients…

but yeah, for now "bats runs installer" sounds OK enough

ehelms:
> if only we had a way to test different server types, similarly we can for clients…

Is that implying a dedicated pipeline?

ekohl:
I'm not sure if a dedicated pipeline makes sense. You also want to test toggling some features. Thinking out loud, you can think of:
Vanilla -> Add REX -> Add Katello -> Enable Katello agent -> Disable Katello agent. Today it's not practical, but I can imagine that's what users would like to do.

ehelms:
Noting the bats update that would go with this change -- https://github.com/theforeman/forklift/pull/1331

evgeni:
> > if only we had a way to test different server types, similarly we can for clients…
> 
> Is that implying a dedicated pipeline?

That would imply it today, but I don't think it's worth one.

It'd be cool if we could setup a system and then run multiple copies of it (snapshot/COW style). But we don't have that today.

manifests/qpid.pp
```diff
@@ -10,31 +10,50 @@
...
+    $ssl_cert_name = $certs::qpid::nss_cert_name
+
+    Class['certs', 'certs::qpid'] ~> Class['qpid']
+  } else {
+    $_qpid_ensure = 'absent'
+    $ssl = false
+    $ssl_cert_db = undef
+    $ssl_cert_password_file = undef
+    $ssl_cert_name = undef
+  }
```
ehelms, line 34: Introduced only including and configuring certificates when katello agent is enabled. Without this when turning it off, the `qpidd` group would not be present and cause the certificate setup to fail. This was caught by the acceptance tests.

manifests/qpid.pp
```diff
@@ -10,31 +10,50 @@
   String $interface = 'lo',
   Integer[0, 5000] $wcache_page_size = 4,
 ) {
-  include certs
-  include certs::qpid
   include katello::params
 
   $agent_event_queue_name = $katello::params::agent_event_queue_name
```
wbclark, line 15: NIT: I think this should be moved under the if block (the one further down at line 50) where it is used

spec/acceptance/qpid_spec.rb
```diff
@@ -1,20 +1,59 @@
...
-  describe port('5671') do
-    it { is_expected.to be_listening }
+    describe port('5672') do
+      it { is_expected.to be_listening }
+    end
   end
 
-  describe port('5672') do
-    it { is_expected.to be_listening }
+  context 'with enable_katello_agent true' do
```
wbclark, line 30:
```suggestion
  context 'with enable_katello_agent false' do
```

spec/classes/qpid_spec.rb
```diff
@@ -5,16 +5,25 @@
     context "on #{os}" do
       let(:facts) { facts }
 
+
```
wbclark, line 8:
```suggestion
```

The extra line here looks unintended.

spec/classes/qpid_spec.rb
```diff
@@ -29,6 +38,31 @@
...
+            .with_ssl(false)
+            .with_ssl_cert_db(nil)
+            .with_ssl_cert_password_file(nil)
+            .with_ssl_cert_name(nil)
+        end
+
+        it do
+          is_expected.not_to create_qpid__config__queue('katello.agent')
+            .with_ssl_cert('/etc/pki/katello/certs/foo.example.com-qpid-broker.crt')
+            .with_ssl_key('/etc/pki/katello/private/foo.example.com-qpid-broker.key')
```
wbclark, line 63: What is actually the behavior of

```
it do
  is_expected.not_to create_...('...')
    .with_...('...')
end
```
?

Intuitively it seems like adding `.with_...` wouldn't change much for the test. Is it sharper coverage to remove the `.with_...` and instead test that the cert/key pair itself doesn't get created?

manifests/qpid.pp
ehelms: This is used inside the `qpid_acls.acl` template.

metadata.json
```diff
@@ -30,7 +30,7 @@
     },
     {
       "name": "katello/qpid",
-      "version_requirement": ">= 7.0.0 < 8.0.0"
+      "version_requirement": ">= 7.1.0 < 8.0.0"
```
ekohl, line 33: 7.1.0 doesn't contain the patches needed. Making a note here so we don't forget to update it in the final version.

spec/classes/qpid_spec.rb
```diff
@@ -6,15 +6,23 @@
       let(:facts) { facts }
 
       context 'with default parameters' do
+        let(:pre_condition) do
+          <<-PUPPET
+          class { 'katello::globals':
+            enable_katello_agent => true,
+          }
+          PUPPET
+        end
```
ekohl, line 15: This is at odds: default parameters with a block that overrides parameters.

spec/classes/qpid_spec.rb
```diff
@@ -29,6 +37,27 @@
...
       end
+
+      context 'with enable_katello_agent false' do
+        let(:pre_condition) do
+          <<-PUPPET
+          class { 'katello::globals':
+            enable_katello_agent => false,
+          }
+          PUPPET
+        end
```
ekohl, line 48: This is actually the default.

ekohl (commented):
This does look correct to me. Only the tests should match their descriptions and we need to ensure a release of qpid exists with the right version number to require.

manifests/params.pp
```diff
@@ -35,5 +35,13 @@
   Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:${candlepin_port}/candlepin",
   String[1] $candlepin_client_keypair_group = 'foreman',
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
+  Variant[Boolean, Enum['']] $enable_katello_agent = false,
```
ekohl, line 38: This could use a `@param`. I'd also explain that the empty string is due to Hiera. That kind of knowledge tends to be useful for people who haven't played with it and find it odd.

manifests/params.pp
```diff
@@ -35,5 +35,13 @@
...
 ) inherits katello::globals {
+
+  if $enable_katello_agent == '' {
+    $enable_katello_agent_real = false
+  } else {
+    $enable_katello_agent_real = $enable_katello_agent
+  }
```
ekohl, line 45: Wouldn't this also work?
```suggestion
  $enable_katello_agent_real = $enable_katello_agent == true
```

ekohl (commented):
Just writing out thoughts. This does mean that an installer user who disables the content proxy bits can't enable Katello Agent, but I don't think that's really a scenario we support today. Those who do need that, can override it via custom-hiera.yaml.

Small notes on the params.pp, but overall :+1:

ehelms:
In some doing testing of repeated enable and disables I came across the needed to manage the Qpid data directory. Therefore https://github.com/theforeman/puppet-qpid/pull/161 should be the last fix needed to do a puppet-qpid release and then update the metadata here with that version of puppet-qpid.

spec/classes/application_spec.rb
```diff
@@ -280,6 +286,96 @@ class {'katello::globals':
...
+            '  :katello_applicability: true',
+          ]
+        end
+
+        it 'should generate correct katello.yaml' do
+          verify_exact_contents(catalogue, '/etc/foreman/plugins/katello.yaml', katello_yaml_content)
+        end
+      end
+
+      context 'with katello::params::enable_katello_agent empty string' do
```
wbclark, line 335: @ehelms could you explain why this empty string test is needed?

wbclark: So this becomes `$enable_katello_agent_real => false`. Ok :+1:

wbclark: And `katello::params::enable_katello_agent` specifically supports the data types `Variant[Boolean, Enum['']]`

wbclark: So what is the benefit of allowing both `$enable_katello_agent => false` and `$enable_katello_agent => ''` and making them equivalent?

spec/classes/application_spec.rb
```diff
@@ -280,6 +286,96 @@ class {'katello::globals':
...
+            '    :oauth_secret: "candlepin-secret"',
+            '    :ca_cert_file: /etc/pki/katello/certs/katello-default-ca.crt',
+            '  :candlepin_events:',
+            '    :ssl_cert_file: /etc/pki/katello/certs/java-client.crt',
+            '    :ssl_key_file: /etc/pki/katello/private/java-client.key',
+            '    :ssl_ca_file: /etc/pki/katello/certs/katello-default-ca.crt',
+            '  :agent:',
+            '    :enabled: false',
+            '    :broker_url: amqps://localhost:5671',
+            '    :event_queue_name: katello.agent',
```
wbclark, line 370:
```suggestion
            '  :agent:',
            '    :enabled: false',
```

Should we simplify the config and omit these when the katello agent is disabled?

wbclark: I think it would be straightforward to wrap those lines with a conditional in templates/katello.yaml.erb

spec/classes/application_spec.rb
ehelms: The explanation I put into the params doc is:

```
# @param enable_katello_agent
#   Set to true to setup Qpid and katello-agent infrastructure. Accepts
#   empty string value to allow Hiera configuration of the value.
```

Does that answer your question or could I word this differently to convey it better?

spec/classes/application_spec.rb
ehelms: Good idea, done.

ekohl (approved):
This looks good to me. I'll leave it up to you to merge it since there is probably some coordination of PRs needed.

ekohl (approved):

PR #388, opened 2021-02-20
Refs #31878 - Allow external mechanism for qpid SASL

ehelms:
On EL8, the SASL enforcement is more pronounced and does not allow anonymous connections when using qpid-stat or qpid-config. We need to specify the mechanism, external for SSL client certificates, and the user, CN of the certificates being used, to call the command. This shows how we can modify this single call we used to do this. The alternative is we invest more in puppet-qpid and add these options directly and then do a release of puppet-qpid.

### DIFF
--- a/spec/acceptance/qpid_spec.rb
+++ b/spec/acceptance/qpid_spec.rb
@@ -1,0 +1,20 @@
+require 'spec_helper_acceptance'
+
+describe 'Install Candlepin' do
+  let(:pp) { 'include katello::qpid' }
+
+  it_behaves_like 'a idempotent resource'
+
+  describe service('qpidd') do
+    it { is_expected.to be_running }
+    it { is_expected.to be_enabled }
+  end
+
+  describe port('5671') do
+    it { is_expected.to be_listening }
+  end
+
+  describe port('5672') do
+    it { is_expected.to be_listening }
+  end
+end
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -12,6 +12,12 @@ group { 'foreman':
   ensure => present,
 }
 
+if $facts['os']['release']['major'] == '7' {
+  package { 'epel-release':
+    ensure => installed,
+  }
+}
+
 if $facts['os']['release']['major'] == '8' {
   # https://tickets.puppetlabs.com/browse/PUP-9978
   exec { '/usr/bin/dnf -y module enable pki-core':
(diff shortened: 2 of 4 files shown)


spec/setup_acceptance_node.pp
```diff
@@ -20,4 +26,17 @@
   package { 'glibc-langpack-en':
     ensure => installed,
   }
+
+  yumrepo { 'katello_staging':
```
ehelms, line 30: Required work around since these packages needed are not in the nightly repository just yet, a chicken and egg problem where these tests need nightly to get released but nightly needs the fixes here to release. I added the tests and caveats to show that this change does work on El8.

evgeni:
Just for my understanding, this is fixing the following issue we see on EL8 nightly right now:
```
 [ERROR ] [configure] 'qpid-config --ssl-certificate /etc/pki/katello/certs/pipe-katello-server-nightly-centos8.n56.example.com-qpid-broker.crt --ssl-key /etc/pki/katello/private/pipe-katello-server-nightly-centos8.n56.example.com-qpid-broker.key -b amqps://localhost:5671 add queue katello.agent --durable' returned 1 instead of one of [0]
```

And the Katello repo you're adding is to get the qpid bits we've built?

I think this is reasonable and we can drop the `yumrepo` once nightly is green.

evgeni:
And the alternative is to either add `sasl=EXTERNAL` to `qpid::config::queue` or allow consumers to pass arbitrary cmd args to `qpid::config::queue` and use that here to pass `sasl=EXTERNAL` or allow to pass a `sasl` arg or ….

That *does* sound cleaner to me, but I am not sure the additional work is worth it. Given the lifetime of qpid in our stack is limited, I'd lean towards the cheapest fix, which this PR implements.

spec/setup_acceptance_node.pp
```diff
@@ -12,6 +12,12 @@
...
 
+if $facts['os']['release']['major'] == '7' {
+  package { 'epel-release':
+    ensure => installed,
+  }
+}
+
 if $facts['os']['release']['major'] == '8' {
   # https://tickets.puppetlabs.com/browse/PUP-9978
   exec { '/usr/bin/dnf -y module enable pki-core':
```
ekohl, line 23: Unrelated, but now that we've fixed it in candlepin, we can actually remove this

spec/setup_acceptance_node.pp
```diff
@@ -20,4 +26,17 @@
...
+    baseurl  => "http://koji.katello.org/releases/yum/katello-nightly/katello/el8/x86_64/",
+    enabled  => true,
+    gpgcheck => false,
+  }
+
+  package { 'dnf-plugins-core':
+    ensure => installed,
+  } ~>
+  exec { '/usr/bin/dnf config-manager --set-enabled powertools':
+  }
```
ekohl, line 41:
```suggestion
  yumrepo { 'powertools':
    enabled => true,
  }
```

spec/acceptance/qpid_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper_acceptance'
+
+describe 'Install Candlepin' do
```
ekohl, line 3:
```suggestion
describe 'Install qpid' do
```

ekohl (commented):
I'll take a look at expanding `qpid::config::queue` with SASL properties.

PR #389, opened 2021-02-24
Update Github Actions workflow

ekohl:
This applies the lessons learned while applying this to Voxpupuli and The Foreman orgs.

Also cleans up custom facts in spec tests.


ekohl:
Puppet 5 is broken since the git version of camptocamp/systemd is no longer Puppet 5 compatible.

PR #398, opened 2021-04-15
Update Limitations in README

### LINKED ISSUE #397: Question: RHEL 8 support
As noted in the README as of now RHEL 8 is not supported.
Is this a hard limit or is this just something where it wasn't properly tested on RHEL 8?

We are in the middle of setting up and are somewhat set on RHEL 8.
Would've been a bummer if we had to spin up a single RHEL 7 node for something small.

moritzdietz:
Update the supported platforms to include RHEL8 as well. Support for this was added with #335 but this change didn't update the README as well.

Closes #397

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This module is designed to setup a full katello server, including candlepin, pul
 
 ## Limitations
 
-* Supports EL7 (RHEL7 / CentOS 7) and EL8 (RHEL8 / CentOS 8)
+* OS Support and other requirements should be checked in the [metadata.json](metadata.json)
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This module is designed to setup a full katello server, including candlepin, pul
 
 ## Limitations
 
-* EL7 (RHEL7 / CentOS 7)
+* Supports EL7 (RHEL7 / CentOS 7) and EL8 (RHEL8 / CentOS 8)
 
 ## Development
 


moritzdietz:
I wasn't sure how you'd like the wording / listing of supported platforms.
Let me know if this should be different in case there are preferences otherwise.

README.md
```diff
@@ -19,7 +19,7 @@ This module is designed to setup a full katello server, including candlepin, pul
 
 ## Limitations
 
-* EL7 (RHEL7 / CentOS 7)
+* Supports EL7 (RHEL7 / CentOS 7) and EL8 (RHEL8 / CentOS 8)
```
ekohl, line 22: How about we refer users to `operatingsystem_support` in `metadata.json` for the exact OS support?

ekohl: > Pro: one less thing to update whenever the supported OSes change (but how often does that _really_ happen)

This is what I'm mostly after. As you can see, it doesn't happen enough that we tend to mess it up. Especially since not all modules have it in their README. However, our CI relies on metadata.json so that's guaranteed to be correct. So it would be my preference to refer users there.

moritzdietz: I've just pushed another commit with an updated text.

ekohl (approved):
I'm not sure how well links render on the forge but I'm ok with it.

ekohl:
Thanks!

moritzdietz:
Ah I would've squashed those but thanks for doing it.
Huh yes i haven't thought about the forge so valid point.
Just thought I'd somehow link it.

evgeni:
Doesn't the forge display the metadata anyways?

moritzdietz:
> Doesn't the forge display the metadata anyways?
Yes, it renders the contents of the metadata so that it's visible there which platform to use.
But if folks browse the GitHub repository it won't be any help there. Having the readme either specify it directly or at least link to info regarding this helps all users wherever they look at the source code. But that's just IMHO.

PR #399, opened 2021-04-15
Remove Scientific 8 support from metadata.json

ekohl:
There is no ScientificLinux 8. The effort was folded into the CentOS project.


evgeni (approved):

wbclark:
I'm labeling this 'skip-changelog' to avoid the potential false impression of changelog readers that there is a scientific linux 8 and we are removing support for it

PR #401, opened 2021-04-20
Fixes #32160 - enable ansible collection content

jlsherrill:


jlsherrill:
Note that we want to get rid of these content_type options, and its on our board to do, but there may be some technical issues that prevent this or make it difficult, we'll figure that out

ekohl:
@ehelms what are you thoughts on refactoring all content parameters to params.pp like we did with qpid? If they're going away eventually, we now know a technique that allows to get rid of the parameters today.

ehelms:
Here is that idea in action for existing types -- https://github.com/theforeman/foreman-installer/pull/676

We let puppet-foreman_proxy_content parameters drive the configuration for Katello. This allows users to have one setting to set and to prevent a discontinuity between what's enabled.

manifests/init.pp
```diff
@@ -58,6 +60,7 @@
   Boolean $enable_file = true,
   Boolean $enable_docker = true,
   Boolean $enable_deb = true,
+  Boolean $enable_ansible_collection = true,
```
ehelms, line 63: If we go this route (https://github.com/theforeman/puppet-katello/pull/402) this file wont need any changes.

manifests/globals.pp
```diff
@@ -12,11 +12,15 @@
...
 #
+# @param enable_ansible_collection
+#   Enable ansible collection content plugin
+#
 class katello::globals(
   Boolean $enable_yum = true,
   Boolean $enable_file = true,
   Boolean $enable_docker = true,
   Boolean $enable_deb = true,
+  Boolean $enable_ansible_collection = true,
```
ehelms, line 23:
```suggestion
    Katello::HieraBoolean $enable_ansible_collection = true,
```

jlsherrill:
i think i updated this correctly?

ekohl (approved):
That looks good (together with https://github.com/theforeman/foreman-installer/pull/677)

manifests/globals.pp
```diff
@@ -12,11 +12,15 @@
 # @param enable_deb
 #   Enable debian content plugin
 #
+# @param enable_ansible_collection
+#   Enable ansible collection content plugin
```
wbclark, line 16:
```suggestion
#   Enable ansible collection content plugin to sync and serve ansible content
```

Since we are also promoting ansible collections to manage Foreman/Katello config and operations, this wording will avoid potential user confusion

ekohl: Technically this goes for all these content plugin parts. Today they don't really enable any plugin but really configure Katello to enable/disable content handling. Perhaps we should have a follow up PR to clean up the phrasing? I haven't really looked too carefully at it since it's on the roadmap to remove (some technical challenges blocking it now) and now we've made sure that this isn't actually exposed to installer users.

ehelms:
Before we merge, I'd like to ensure we have an answer to https://github.com/theforeman/puppet-foreman_proxy_content/pull/343/files#r617586412

manifests/globals.pp
wbclark: OK, thanks for the explanation. A follow up PR to clean up param phrasing sounds good to me. I'm sure there are other places in the installer module ecosystem where we would benefit as well, including some that are exposed to installer users.

ehelms:
You'll need a rebase now that I broke your PR merging mine :)

jlsherrill:
errors look like related to setting up tomcat?

ehelms:
Yea, unrelated to this. Fixed in puppet-candlepin directly, going to re-run just to be safe.

ehelms:
This is failing due to an odd failure that has popped up and can be seen more directly here -- https://github.com/theforeman/puppet-certs/pull/323

Given that, and that this change does not affect that area I am going to merge the changes related to this.

PR #402, opened 2021-04-20
Refs #32369 - Move content enable flags to params

ehelms:
This moves the content enable flags to the params class and away from
init.pp. This will allow a user to either set these via the globals
class or via hiera.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -13,10 +13,10 @@
 #   Enable debian content plugin
 #
 class katello::globals(
-  Boolean $enable_yum = true,
-  Boolean $enable_file = true,
-  Boolean $enable_docker = true,
-  Boolean $enable_deb = true,
+  Katello::HieraBoolean $enable_yum = true,
+  Katello::HieraBoolean $enable_file = true,
+  Katello::HieraBoolean $enable_docker = true,
+  Katello::HieraBoolean $enable_deb = true,
 ) {
   # OAUTH settings
   $candlepin_oauth_key = 'katello'
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -36,4 +36,10 @@ class katello::params (
   String[1] $candlepin_client_keypair_group = 'foreman',
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
 ) inherits katello::globals {
+
+  $enable_yum = $katello::globals::enable_yum != false
+  $enable_file = $katello::globals::enable_file != false
+  $enable_docker = $katello::globals::enable_docker != false
+  $enable_deb = $katello::globals::enable_deb != false
+
 }
--- a/spec/classes/application_spec.rb
+++ b/spec/classes/application_spec.rb
@@ -280,6 +280,77 @@ describe 'katello::application' do
           it { is_expected.to create_package('rubygem-katello').that_requires('Anchor[katello::candlepin]') }
         end
       end
+
+      context 'with katello::globals parameter as an empty string' do
+        let :pre_condition do
+          <<-EOS
+          class {'katello::globals':
+            enable_deb => '',
+          }
+          #{super()}
+          EOS
+        end
+
+        it { is_expected.to compile.with_all_deps }
+
+        let(:katello_yaml_content) do
+          if facts[:operatingsystemmajrelease] == '7'
+            [
+              ':katello:',
+              '  :rest_client_timeout: 3600',
+              '  :content_types:',
+              '    :yum: true',
+              '    :file: true',
+              '    :deb: true',
+              '    :puppet: false',
+              '    :docker: true',
+              '    :ostree: false',
+              '  :candlepin:',
+              '    :url: https://localhost:23443/candlepin',
+              '    :oauth_key: "katello"',
+              '    :oauth_secret: "candlepin-secret"',
+              '    :ca_cert_file: /etc/pki/katello/certs/katello-default-ca.crt',
+              '  :candlepin_events:',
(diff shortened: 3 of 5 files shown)


manifests/params.pp
```diff
@@ -35,5 +44,9 @@
   Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:${candlepin_port}/candlepin",
   String[1] $candlepin_client_keypair_group = 'foreman',
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
+  Boolean $enable_yum = $katello::globals::enable_yum,
+  Boolean $enable_file = $katello::globals::enable_file,
+  Boolean $enable_docker = $katello::globals::enable_docker,
+  Boolean $enable_deb = $katello::globals::enable_deb,
```
ekohl, line 50: Do we even need these or can we just use the globals directly in the installer? We probably need to accept `Enum['']` there as well since we use `alias()`.

manifests/params.pp
```diff
@@ -24,6 +24,15 @@
 #   The oauth secret for Candlepin
 # @param postgresql_evr_package
 #   The contextual package name for the PostgreSQL EVR extension
+# @param enable_yum
+#   Enable rpm content plugin, including syncing of yum content
+# @param enable_file
+#   Enable generic file content management
+# $enable_docker
```
ekohl, line 31: Unrelated, but we should rename this to `container` at some point.

ekohl (changes requested):

ehelms:
Switched to just using the globals values to set these parameters, reflected here to -- https://github.com/theforeman/foreman-installer/pull/676

ekohl (approved):
This should work since in Puppet 4+ an empty string is considered true. It doesn't hurt to have a test case for that though. Otherwise :+1:

ehelms:
So passing an empty string results in:

```
       the extra elements were:        ["    :deb: "]
```

So I think that means I either need to do a conversion using a "real" variable ?

ekohl:
There's 3 options that I see:
* Use `_real` or `real_` in `globals.pp`
* Use `$enable_deb = $katello::globals::enable_deb != false` in `params.pp` (inside the body, not as a class param)
* Use `$enable_deb ? 'true' : 'false'` in the template

I'm leaning to option 2. `application.pp` already uses the variable from `katello::params` so it means you don't need to make any other modifications and it can be treated as a real boolean then.

It may make sense to introduce a function
```puppet
# @summary Normalize a boolean that also accepts an empty string
# @return A normalized boolean. Empty strings are considered true
# @api private
function katello::normalize_boolean(Variant[Boolean, Enum['']] $value) >> Boolean {
  $value != false
}
```

Now that I write it down, perhaps `Boolean()` instead of a custom function works just as well.

Given the repetition, perhaps a type alias is also useful. That can avoid having to repeat the explanation in the parameter documentation.
```puppet
# @summary A boolean that also accepts empty strings
# In Hiera the function alias() returns an empty string if the value is undefined.
type Katello::HieraBoolean = Variant[Boolean, Enum['']]
```

spec/classes/application_spec.rb
```diff
@@ -280,6 +280,77 @@ class {'katello::globals':
...
+              '  :agent:',
+              '    :broker_url: amqps://localhost:5671',
+              '    :event_queue_name: katello.agent',
+              '  :katello_applicability: true',
+            ]
+          end
+        end
+
+        it 'should generate correct katello.yaml' do
+          verify_exact_contents(catalogue, '/etc/foreman/plugins/katello.yaml', katello_yaml_content)
```
ekohl, line 351: You can also use `verify_contents(catalogue, '/etc/foreman/plugins/katello.yaml', ['    :deb: true'])` to verify a single line.

ehelms:
> Now that I write it down, perhaps `Boolean()` instead of a custom function works just as well.

This did not seem to work:

```
'new_boolean' The string '' cannot be converted to Boolean
```

> * Use `$enable_deb = $katello::globals::enable_deb != false` in `params.pp` (inside the body, not as a class param)

I opted for this strategy as it worked and felt very straight forward using a native mechanism.

spec/classes/application_spec.rb
ehelms: I'm gonna do a follow up PR to fix up some of the tests and will tackle this. Cause I noticed there is some EL7 vs. EL8 tests which are no longer valid (as we have no EL7 caveats anymore).

manifests/globals.pp
```diff
@@ -2,21 +2,29 @@
 #
 # @param enable_yum
 #   Enable rpm content plugin, including syncing of yum content
+#   Accepts empty string value, the default value when no lookup
+#   is found, to allow Hiera configuration of the value.
```
ekohl, line 6: With the type alias, I think these comments can be dropped.

ehelms: Ah yea, thanks for catching that. I got too excited and pushed early.

ekohl:
> This did not seem to work:

I could have checked it. Another thing that came to my mind is that perhaps not everything will interpret an empty string as true. It's more like `undef` and our defaults happen to be `true` now. So I'm fine with being explicit and not too smart.

spec/classes/application_spec.rb
ekohl: Not having EL7 caveats makes me happy.

ehelms:
@ekohl Can you review https://github.com/theforeman/foreman-installer/pull/676 before I merge this change

ekohl (approved):

PR #403, opened 2021-04-21
Clean up some of our tests

ehelms:


spec/classes/application_spec.rb
```diff
@@ -140,61 +113,13 @@ class { 'katello::params':
...
-            ]
-          end
+          [
+            '  :rest_client_timeout: 4000',
+          ]
         end
 
         it 'should generate correct katello.yaml' do
-          verify_exact_contents(catalogue, '/etc/foreman/plugins/katello.yaml', katello_yaml_content)
+          verify_contents(catalogue, '/etc/foreman/plugins/katello.yaml', katello_yaml_content)
```
ekohl, line 122: Thinking out loud: does it make sense to shorten this and drop the `let` block?
```suggestion
          verify_contents(catalogue, '/etc/foreman/plugins/katello.yaml', ['  :rest_client_timeout: 4000'])
```

On the other hand, it keeps the tests consistent since they all follow the same pattern.

ekohl (approved):

spec/classes/application_spec.rb
ehelms: Yea, I both liked the pattern nature of the tests and the fact that it 'looks' like yaml which I find as a person writing or reading the tests easier to translate to a deployment.

ekohl (approved):
The EL8 failure looks like Candlepin which we also saw in puppet-candlepin. Also can't be related to these changes.

PR #405, opened 2021-04-22
Refs #32381: Add function to determine the Artemis user string dynamically from th…

ehelms:
…e certificate

This suffers from this issue: https://github.com/theforeman/puppet-katello/pull/404
This is not intended for the upcoming Foreman 2.5 release

This value is currently declared in puppet-certs in a hard-coded manner (https://github.com/theforeman/puppet-certs/blob/master/manifests/candlepin.pp#L29). This is a fragile method. By deriving this directly from the certificate used, we can instead more safely modify and change the certificates.

This is the first part of a change I intend to make that moves away from using `certs::candlepin::client_cert` and moves instead to using `certs::foreman::client_cert` which will help reduce the number of certificates generated and the number of different certificates configured for use of inside Foreman. This aims towards that single set of client certificates for Foreman used to talk to our different backends goal.

### DIFF
--- a/lib/puppet/functions/katello/certificate_subject.rb
+++ b/lib/puppet/functions/katello/certificate_subject.rb
@@ -1,0 +1,23 @@
+# @summary
+#   Extracts a certificate's subject in RFC2253 form
+#
+# @example How to extract a certificate's subject
+#   $cert_subject = katello::certificate_subject($path_to_certificate)
+#
+Puppet::Functions.create_function(:'katello::certificate_subject') do
+  # @param certificate_path
+  #
+  # @return [String]
+  dispatch :certificate_subject do
+    param 'Stdlib::Absolutepath', :certificate_path
+  end
+
+  def certificate_subject(certificate_path)
+    begin
+      cert = OpenSSL::X509::Certificate.new(File.read(certificate_path))
+      cert.subject.to_s(OpenSSL::X509::Name::RFC2253)
+    rescue OpenSSL::X509::CertificateError, Errno::ENOENT
+      nil
+    end
+  end
+end
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -35,6 +35,8 @@ class katello::candlepin (
     client_keypair_group => $katello::params::candlepin_client_keypair_group,
   }
 
+  $artemis_client_dn = katello::certificate_subject($certs::candlepin::client_cert)
+
   class { 'candlepin':
     host                         => $katello::params::candlepin_host,
     ssl_port                     => $katello::params::candlepin_port,
@@ -47,7 +49,7 @@ class katello::candlepin (
     keystore_password            => $certs::candlepin::keystore_password,
     truststore_file              => $certs::candlepin::truststore,
     truststore_password          => $certs::candlepin::truststore_password,
-    artemis_client_dn            => $certs::candlepin::artemis_client_dn,
+    artemis_client_dn            => $artemis_client_dn,
     java_home                    => '/usr/lib/jvm/jre-11',
     java_package                 => 'java-11-openjdk',
     enable_basic_auth            => false,
--- a/spec/functions/katello_certificate_subject_spec.rb
+++ b/spec/functions/katello_certificate_subject_spec.rb
@@ -1,0 +1,50 @@
+require 'spec_helper'
+
+describe 'katello::certificate_subject' do
+  let(:test_cert) do
+  <<-CERT
+-----BEGIN CERTIFICATE-----
+MIIC+TCCAeGgAwIBAgIUetO+zvwJ4nLNrxe9lcrT4h0noCMwDQYJKoZIhvcNAQEL
+BQAwHjEcMBoGA1UEAwwTVGVzdCBTZWxmLVNpZ25lZCBDQTAeFw0yMDExMTgwMjMw
+NDNaFw0zMDExMTYwMjMwNDNaMB4xHDAaBgNVBAMME2ZvcmVtYW4uZXhhbXBsZS5j
+b20wggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDcEtPd1uKX0hct7+qe
+gOEy72VB93cBGuEJis6yD7uJfdjnbBtiFwkxUqmQlsDmUsqcuh6106yDkaW6tyzT
+I6R0Xx8OJkT4bxOsgkr3xqZSrAQJvn/NmV4j6egckJlgYnSbkrOFvy5iO1A/Dc/m
+OrC6TJVGe/YvMCU6IYPU1f/acNucRZGopa7yfhyTd8nzArq1BCSrjqtCl8m9NPJZ
+IP8+06wQ6MCjyd+kjnm+Tq/P+mKEsXVDBQCQAyWFpZdUcu4zbL+UV2+O7QUtndEh
+k2nf4w3Rx70XvMwagfo3hE5cJ8rNXEynphhDzdJqzRDpPYItZauMDxmK+4oHOn0g
+90t5AgMBAAGjLzAtMAsGA1UdDwQEAwIFIDAeBgNVHREEFzAVghNmb3JlbWFuLmV4
+YW1wbGUuY29tMA0GCSqGSIb3DQEBCwUAA4IBAQBlIEA2CfZIa8LtUYlDwa5v+5Wf
+1ktmYTRtgEI+922T/eTB8uH1//VxpfK5ynljao7SNVcX+74Q+YH/4Ci4OfZvE5vA
+1IJXog5bfE4mVc1qXhH7TokBQx1L6vtUh9OaTGpBAVnS3J5jLw6+Tdi9FOeZdKHZ
+FvMnyZ7MQ6VjbLZsTy49o87Nstqkle48ivwSFrDU1cDN+6S/DUdHQnh8XtPB1PMh
+7WCxGGtzmw5s5SxBkyY/buGDr+kx52yULl6ZrnJD6PfR30X+8G3ltvmaCQllYadX
+eprUs5H2WDnTUUE78+cf1JK29Zs9it/l4t2uLc5Z94oXosFLkTKw6ZSB3X9J
+-----END CERTIFICATE-----
+  CERT
+  end
+
+  it 'should exist' do
+    is_expected.not_to eq(nil)
+  end
+
+  it 'should return a certificate subject in RFC2253 format' do
+    file = class_double("File")
+    allow(File).to receive(:read).and_call_original
+    allow(File).to receive(:read).with("/tmp/client_cert.crt").and_return(test_cert)
+
+    is_expected.to run.with_params('/tmp/client_cert.crt').and_return("CN=foreman.example.com")
+  end
+
+  it 'should handle a bad certificate' do
+    file = class_double("File")
+    allow(File).to receive(:read).and_call_original
+    allow(File).to receive(:read).with("/tmp/client_cert.crt").and_return('aab32433adfad')
+
+    is_expected.to run.with_params('/tmp/client_cert.crt').and_return(nil)
+  end
+
+  it 'should handle a non-existent file' do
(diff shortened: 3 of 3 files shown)


lib/puppet/functions/katello/certificate_subject.rb
```diff
@@ -0,0 +1,24 @@
+# @summary
+#   Extracts a certificate's subject in RFC2253 form
+#
+# @example How to extract a certificate's subject
+#   $cert_subject = katello::certificate_subject($path_to_certificate)
+#
+Puppet::Functions.create_function(:'katello::certificate_subject') do
+  # @param certificate_path
+  # @param options
```
ekohl, line 9: I don't see this param.
```suggestion
```

lib/puppet/functions/katello/certificate_subject.rb
```diff
@@ -0,0 +1,24 @@
...
+  #
+  # @return [String]
+  dispatch :certificate_subject do
+    param 'Any', :certificate_path
```
ekohl, line 13: Should this be an absolute path?
```suggestion
    param 'Stdlib::Absolutepath', :certificate_path
```

lib/puppet/functions/katello/certificate_subject.rb
```diff
@@ -0,0 +1,24 @@
+# @summary
+#   Extracts a certificate's subject in RFC2253 form
+#
+# @example How to extract a certificate's subject
+#   $cert_subject = katello::certificate_subject($path_to_certificate)
+#
+Puppet::Functions.create_function(:'katello::certificate_subject') do
```
ekohl, line 7: I wonder if this should live in the `certs` module as `certs::subject`.

manifests/candlepin.pp
```diff
@@ -47,7 +47,7 @@
     keystore_password            => $certs::candlepin::keystore_password,
     truststore_file              => $certs::candlepin::truststore,
     truststore_password          => $certs::candlepin::truststore_password,
-    artemis_client_dn            => $certs::candlepin::artemis_client_dn,
+    artemis_client_dn            => katello::certificate_subject($certs::candlepin::client_cert),
```
ekohl, line 50: I'm not sure if this will work in a client-server mode. Functions run server side, but the certificates live client side. I think you need to use a [deferred function](https://puppet.com/docs/puppet/7.5/deferring_functions.html). I think they're relatively recent (we may need to bump our lower version bound) and initial implementations had some issues I've heard. No personal experience with them.

lib/puppet/functions/katello/certificate_subject.rb
```diff
@@ -0,0 +1,24 @@
...
+  # @param options
+  #
+  # @return [String]
+  dispatch :certificate_subject do
+    param 'Any', :certificate_path
+  end
+
+  def certificate_subject(certificate_path)
+    if File.exist?(certificate_path)
+      cert = OpenSSL::X509::Certificate.new(File.read(certificate_path))
```
ekohl, line 18: Does it need a rescue? For example, when the certificate is invalid.

ekohl (commented):
I think this could be a good direction. The benefit is that in theory we could use any CA provider.

This is more future direction thinking, but IMHO we can't expect the client cert and server to be present on the same host. Foreman will have the client cert and Candlepin may (hopefully) run on a different host. That means that `candlepin.pp` should have a parameter for the `artemis_client_dn` and we can determine it in `init.pp` by extracting. However, that's out of scope for this PR.

lib/puppet/functions/katello/certificate_subject.rb
ekohl: Oh, I now read the note:
>  Deferred functions can only use types that are built into Puppet (for example String). They cannot use types from modules like stdlib because Puppet does not plugin-sync these types to the agent. 

Perhaps it's better to use a non-empty string instead?
```suggestion
    param 'String[1]', :certificate_path
```

manifests/candlepin.pp
ehelms: Looks like that would work, from some early testing I would have to update puppet-candlepin so that this parameter accepts a Deferred:

```
Class[Candlepin]: parameter 'artemis_client_dn' expects a String value, got Deferred
```

lib/puppet/functions/katello/certificate_subject.rb
```diff
@@ -0,0 +1,23 @@
...
+#
+# @example How to extract a certificate's subject
+#   $cert_subject = katello::certificate_subject($path_to_certificate)
+#
+Puppet::Functions.create_function(:'katello::certificate_subject') do
+  # @param certificate_path
+  #
+  # @return [String]
+  dispatch :certificate_subject do
+    param 'Stdlib::Absolutepath', :certificate_path
```
ekohl, line 12: It can return `nil`. It's also better to really use a return type.

And I also made a note that functions used in Deferred can't use custom types but only built in ones.

```suggestion
  dispatch :certificate_subject do
    param 'String[1]', :certificate_path
    return_type 'Optional[String]'
```

lib/puppet/functions/katello/certificate_subject.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def certificate_subject(certificate_path)
+    begin
+      cert = OpenSSL::X509::Certificate.new(File.read(certificate_path))
+      cert.subject.to_s(OpenSSL::X509::Name::RFC2253)
+    rescue OpenSSL::X509::CertificateError, Errno::ENOENT
```
ekohl, line 19: Can use a `Puppet.warn` I think

spec/functions/katello_certificate_subject_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+k2nf4w3Rx70XvMwagfo3hE5cJ8rNXEynphhDzdJqzRDpPYItZauMDxmK+4oHOn0g
+90t5AgMBAAGjLzAtMAsGA1UdDwQEAwIFIDAeBgNVHREEFzAVghNmb3JlbWFuLmV4
+YW1wbGUuY29tMA0GCSqGSIb3DQEBCwUAA4IBAQBlIEA2CfZIa8LtUYlDwa5v+5Wf
+1ktmYTRtgEI+922T/eTB8uH1//VxpfK5ynljao7SNVcX+74Q+YH/4Ci4OfZvE5vA
+1IJXog5bfE4mVc1qXhH7TokBQx1L6vtUh9OaTGpBAVnS3J5jLw6+Tdi9FOeZdKHZ
+FvMnyZ7MQ6VjbLZsTy49o87Nstqkle48ivwSFrDU1cDN+6S/DUdHQnh8XtPB1PMh
+7WCxGGtzmw5s5SxBkyY/buGDr+kx52yULl6ZrnJD6PfR30X+8G3ltvmaCQllYadX
+eprUs5H2WDnTUUE78+cf1JK29Zs9it/l4t2uLc5Z94oXosFLkTKw6ZSB3X9J
+-----END CERTIFICATE-----
+  CERT
```
ekohl, line 24: You can use indenting with `<<~`:
```suggestion
    <<~CERT
      -----BEGIN CERTIFICATE-----
      MIIC+TCCAeGgAwIBAgIUetO+zvwJ4nLNrxe9lcrT4h0noCMwDQYJKoZIhvcNAQEL
      BQAwHjEcMBoGA1UEAwwTVGVzdCBTZWxmLVNpZ25lZCBDQTAeFw0yMDExMTgwMjMw
      NDNaFw0zMDExMTYwMjMwNDNaMB4xHDAaBgNVBAMME2ZvcmVtYW4uZXhhbXBsZS5j
      b20wggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDcEtPd1uKX0hct7+qe
      gOEy72VB93cBGuEJis6yD7uJfdjnbBtiFwkxUqmQlsDmUsqcuh6106yDkaW6tyzT
      I6R0Xx8OJkT4bxOsgkr3xqZSrAQJvn/NmV4j6egckJlgYnSbkrOFvy5iO1A/Dc/m
      OrC6TJVGe/YvMCU6IYPU1f/acNucRZGopa7yfhyTd8nzArq1BCSrjqtCl8m9NPJZ
      IP8+06wQ6MCjyd+kjnm+Tq/P+mKEsXVDBQCQAyWFpZdUcu4zbL+UV2+O7QUtndEh
      k2nf4w3Rx70XvMwagfo3hE5cJ8rNXEynphhDzdJqzRDpPYItZauMDxmK+4oHOn0g
      90t5AgMBAAGjLzAtMAsGA1UdDwQEAwIFIDAeBgNVHREEFzAVghNmb3JlbWFuLmV4
      YW1wbGUuY29tMA0GCSqGSIb3DQEBCwUAA4IBAQBlIEA2CfZIa8LtUYlDwa5v+5Wf
      1ktmYTRtgEI+922T/eTB8uH1//VxpfK5ynljao7SNVcX+74Q+YH/4Ci4OfZvE5vA
      1IJXog5bfE4mVc1qXhH7TokBQx1L6vtUh9OaTGpBAVnS3J5jLw6+Tdi9FOeZdKHZ
      FvMnyZ7MQ6VjbLZsTy49o87Nstqkle48ivwSFrDU1cDN+6S/DUdHQnh8XtPB1PMh
      7WCxGGtzmw5s5SxBkyY/buGDr+kx52yULl6ZrnJD6PfR30X+8G3ltvmaCQllYadX
      eprUs5H2WDnTUUE78+cf1JK29Zs9it/l4t2uLc5Z94oXosFLkTKw6ZSB3X9J
      -----END CERTIFICATE-----
    CERT
```

ehelms:
> This is more future direction thinking, but IMHO we can't expect the client cert and server to be present on the same host. Foreman will have the client cert and Candlepin may (hopefully) run on a different host. That means that `candlepin.pp` should have a parameter for the `artemis_client_dn` and we can determine it in `init.pp` by extracting. However, that's out of scope for this PR.

I am working to formulate my full thoughts to write it all down. One change in tactics I had been contemplating was moving to a model where we pass to puppet-candlepin the tomcat server certificates and any client certificates that should be present in the truststore. The puppet-candlepin module would then take care of managing the correctly formatted (and owned) keystore needed for tomcat, truststore for client communication and extraction of DN for the Artemis users configuration. So effectively this function I am writing here would live in puppet-candlepin instead. And we would have to do less coordination between three different puppet modules with simplified interface of handing the module certificates and letting it control formatting.

ekohl:
I would prefer, if it's possible, to store a CA that has signed certificates in a truststore and pass a DN for the exact certificate to be trusted.

My reasoning is that you can then rotate the client certificate without having to update the truststore as well.

This is also in line with a Smart Proxy where we specify a CA and then a list of trusted hosts.

ehelms:
> I would prefer, if it's possible, to store a CA that has signed certificates in a truststore and pass a DN for the exact certificate to be trusted.

I'll be the first to admit that I question my interpretation of the docs in these areas every single time I read them as to the intent. For Artemis, the docs say:

> trustStorePath

When used on an acceptor this is the path to the server-side SSL key store that holds the keys of all the clients that the server trusts. This is only relevant for an acceptor if you are using 2-way SSL (i.e. mutual authentication).

Which reads to me like, to achieve mTLS we have to put the client certificate into the truststore. The DN that we pass in is used to configure the Artemis authorization layer. So we are achieving mTLS and using RBAC to constrain.

ekohl:
Do you have a link for that? Also, can we reach out to the Candlepin team or Artemis to get more info on it? It could be supported but undocumented or on their roadmap. Perhaps we can even make sure it's on the roadmap.

ehelms:
> Do you have a link for that? Also, can we reach out to the Candlepin team or Artemis to get more info on it? It could be supported but undocumented or on their roadmap. Perhaps we can even make sure it's on the roadmap.

Sorry, what are you wanting a link for? What is the 'it' in this case?

ekohl:
The documentation you linked.

ehelms:
> The documentation you linked.

https://activemq.apache.org/components/artemis/documentation/1.5.2/configuring-transports.html

ehelms:
Closed in favor of -- https://github.com/theforeman/puppet-certs/pull/332

PR #409, opened 2021-04-27
Always declare the Foreman class

ekohl:
Katello has always used this, but set them via Hiera.

This is just a draft to share some example code and open the discussion.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -10,13 +10,24 @@ class katello::application (
   Integer[0] $rest_client_timeout = 3600,
   Integer[0] $hosts_queue_workers = 1,
 ) {
-  include foreman
   include certs
   include certs::apache
   include certs::candlepin
   include certs::foreman
   include certs::pulp_client
   include katello::params
+
+  class { 'foreman':
+    client_ssl_cert  => $certs::foreman::client_cert,
+    client_ssl_key   => $certs::foreman::client_key,
+    client_ssl_ca    => $certs::foreman::ssl_ca_cert,
+    server_ssl_key   => $certs::apache::apache_key,
+    server_ssl_cert  => $certs::apache::apache_cert,
+    server_ssl_chain => $certs::apache::apache_ca_cert, # TODO: is this correct?
+    server_ssl_ca    => $certs::apache::apache_ca_cert, # TODO: is this correct?
+    user_groups      => [],
+    require          => Class['certs::apache', 'certs::foreman'],
+  }
 
   foreman_config_entry { 'pulp_client_cert':
     value          => $certs::pulp_client::client_cert,


ekohl:
@laugmanuel would this cause problems for you?

ehelms:
Would you explain the mechanism by which the inclusion here does not break it being declared in the answers file for the installer? (https://github.com/theforeman/foreman-installer/blob/develop/config/katello-answers.yaml#L14)

ekohl:
The background behind this is how parameter lookup works. https://puppet.com/docs/puppet/7.6/lang_classes.html#lang_class_declare-section_x54_1hk_xhb has the full text. The critical part is:
> When a class is declared with a resource-like declaration, Puppet takes the following actions, in order, for each of the class parameters:
>
> 1. Uses the override value from the declaration, if present.
> 2. Requests a value from the external data source, using the key <class name>::<parameter name>. For example, to get the apache class's version parameter, Puppet searches for apache::version.
> 3. Uses the default value.
> 4. Fails compilation with an error, if no value is found.

So these overrides will never be looked up which means that whatever is in the answers is irrelevant.

The only gotcha is that `class { 'x': }` must come before any `include x`. This is why Kafo has an order. You can also see that the current order will break:
https://github.com/theforeman/foreman-installer/blob/01773ed6238d96bc6108943362fa06eda084d5b4/config/katello.yaml#L20-L27

You can see `foreman` comes before `katello` which will lead to a compile error **if** this patch is merged. Those are some cases where the compile order in Puppet is important.

ehelms:
> So these overrides will never be looked up which means that whatever is in the answers is irrelevant.

I am having a hard time following this. What about this makes the answers file values irrelevant? Does that imply also that no matter what is declared in the answers file it could never override these declared here in this PR?

ekohl:
Correct. Any parameter which is declared in such a way can't be overridden by Hiera lookups. Any parameter that isn't, is still looked up via Hiera. I'm not entirely sure yet if this is a good pattern or not, but I thought code was a good way to start the discussion.

laugmanuel:
> @laugmanuel would this cause problems for you?

Yes, this would break our current setup as we declare `::foreman` ourselves with quite a lot of parameters (about 35) and afterwards declare `katello::application` with some specific parameters for Katello.

This PR would cause a duplicate declaration of `::foreman` which would not be resolvable.

ekohl:
Thanks for confirming that - any thoughts on a way to deal with this? I've thought about a parameter to either include or manage it this way. Would that be an option?

Just another thought: we can also introduce a `foreman::globals` or even `foreman::certificates` where we can set certificate handling.

laugmanuel:
I'm totally fine with a parameter to decide how `::foreman` should be managed - I'm also fine with a different approach.
However, I do not really have insights into best practices for a case like this.
In the end, the user should be able to reuse the Puppet module(s) in a way they want (e.g. define `::foreman` and `::katello::application` separately).

PR #411, opened 2021-04-30
Fixes #32383: Do not set Pulp client certificates

ehelms:
Rely on the default setting from the application for Pulp client
certificates which will be the Foreman client certificates.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -18,18 +18,6 @@ class katello::application (
   include certs::pulp_client
   include katello::params
 
-  foreman_config_entry { 'pulp_client_cert':
-    value          => $certs::pulp_client::client_cert,
-    ignore_missing => false,
-    require        => [Class['certs::pulp_client'], Foreman::Rake['db:seed']],
-  }
-
-  foreman_config_entry { 'pulp_client_key':
-    value          => $certs::pulp_client::client_key,
-    ignore_missing => false,
-    require        => [Class['certs::pulp_client'], Foreman::Rake['db:seed']],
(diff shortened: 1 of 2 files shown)


ekohl:
Is there a related Katello change to make this work that this also depends on?

ehelms:
> Is there a related Katello change to make this work that this also depends on?

https://github.com/Katello/katello/pull/9330

ekohl:
Converted to a draft to indicate there's design discussion happening in the linked Katello PR and this shouldn't be merged just yet.

ehelms:
The design has been settled and https://github.com/Katello/katello/pull/9330 is ready for review. With respect to ordering, should this change go in first since the Katello change drops the settings?

manifests/application.pp
```diff
@@ -15,21 +15,8 @@
   include certs::apache
   include certs::candlepin
   include certs::foreman
-  include certs::pulp_client
```
ehelms, line 18: Making a note to fix this, we'll still need to deploy these certificates for N-1 support.

manifests/application.pp
```diff
@@ -18,18 +18,6 @@
   include certs::pulp_client
   include katello::params
 
-  foreman_config_entry { 'pulp_client_cert':
```
ekohl, line 21: If we don't set these, how does the application still know about them?

ehelms: Checkout https://github.com/Katello/katello/pull/9330 -- we are dropping the settings and handling the paths there.

PR #415, opened 2021-05-27
Allow certs 14.0.0

ehelms:


wbclark (commented):
On CentOS 8, candlepin status API was returning 404 after a successful puppet apply, but that test suite was run 27 days ago.

Latest in CI was all green at https://github.com/theforeman/puppet-katello/runs/2881527510?check_suite_focus=true .

Giving these a re-run and anticipating they will come back green.

ehelms:
Hooray looks like CentOS 8 container has been updated to 8.4 thus fixing our problem

wbclark (approved):

PR #344, opened 2020-05-14
Release 14.0.0

wbclark:

PR #419, opened 2021-06-21
Refs #32511 - Rename certs::qpid::nss_db_password_path

wbclark:
Requires a merge of puppet-certs #344 and subsequent
release of puppet-cers v12.1.0


wbclark:
The original reason to pursue this rename no longer applies so I am closing this.

PR #423, opened 2021-06-25
Add basic acceptance test for the whole module

ehelms:
Giving this a try to see if it can help find issues earlier and identify any problems with some cross module changes that are orchestrated here. This currently fails because nightly fails.

### DIFF
--- a/spec/acceptance/hieradata/common.yaml
+++ b/spec/acceptance/hieradata/common.yaml
@@ -1,0 +1,9 @@
+---
+foreman::client_ssl_ca: /etc/foreman/proxy_ca.pem
+foreman::client_ssl_cert: /etc/foreman/client_cert.pem
+foreman::client_ssl_key: /etc/foreman/client_key.pem
+foreman::server_ssl_ca: /etc/pki/katello/certs/katello-default-ca.crt
+foreman::server_ssl_cert: /etc/pki/katello/certs/katello-apache.crt
+foreman::server_ssl_chain: /etc/pki/katello/certs/katello-server-ca.crt
+foreman::server_ssl_crl: ""
+foreman::server_ssl_key: /etc/pki/katello/private/katello-apache.key
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -16,6 +16,15 @@ file { '/etc/foreman':
   ensure => directory
 }
 
+# Necessary for PostgreSQL EVR extension
+yumrepo { "pulpcore":
+  baseurl  => "http://yum.theforeman.org/pulpcore/3.14/el\$releasever/x86_64/",
+  descr    => "Pulpcore",
+  enabled  => true,
+  gpgcheck => true,
+  gpgkey   => "https://yum.theforeman.org/pulpcore/3.14/GPG-RPM-KEY-pulpcore",
+}
+
 if $facts['os']['release']['major'] == '8' {
   package { 'glibc-langpack-en':
     ensure => installed,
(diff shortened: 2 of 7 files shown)


spec/setup_acceptance_node.pp
```diff
@@ -7,11 +7,19 @@
 class { 'candlepin::repo':
   version => 'nightly',
 }
-
 group { 'foreman':
   ensure => present,
 }
 
+# Necessary for PostgreSQL EVR extension
```
ekohl, line 14: I ran into the same thing with an experiment. It probably isn't worth to pull in `pulpcore` just for `pulpcore::repo`, but it would have the benefit of not having to maintain it in multiple places.

ekohl (commented):
This has been on my wish list for a while.

spec/setup_acceptance_node.pp
ehelms: Yea, thats why I went this route rather than put another dependency on, and figure we will just solve this via dropping the RPM one day.

ekohl: Agreed, we should work to that goal. Another solution may be to move it to foreman-packaging. AFAIK Pulp doesn't actually use it anyway.

spec/setup_acceptance_node.pp
```diff
@@ -7,11 +7,19 @@
...
+yumrepo { "pulpcore":
+  baseurl  => "http://yum.theforeman.org/pulpcore/3.11/el\$releasever/x86_64/",
```
ekohl, line 16: Before we forget: this should be 3.14 now.

ehelms:
This will pass now that nightlies are green -- first needs https://github.com/theforeman/puppet-katello/pull/425

ehelms:
Includes https://github.com/theforeman/puppet-katello/pull/425 for right now to show that tests pass

wbclark:
This will now require a rebase

manifests/init.pp
```diff
@@ -59,7 +59,7 @@
 ) {
 
   package { 'katello':
-    ensure => installed,
+    ensure  => installed,
```
wbclark, line 62:
```suggestion
    ensure => installed,
```

I believe this is a relic of a previous change that is no longer included. Let's simplify the diff / not add unnecessary white space

spec/support/acceptance/purge.rb
```diff
@@ -0,0 +1,4 @@
+def purge_katello
+  on default, 'yum -y remove foreman* tfm-*'
+  on default, "systemctl stop httpd", { :acceptable_exit_codes => [0, 5] }
+end
```
wbclark, line 4: For completeness, should this drop DBs in postgres as well?

ehelms: Fair question, I borrowed this originally from https://github.com/theforeman/puppet-foreman/blob/master/spec/support/acceptance/purge.rb which did not. To what degree we need to clean up is unclear (Redis? Certs?)

ehelms:
What else would folks like to see here? I think this adds a lot of value and can be built upon so I dont feel we need to go for completeness right away.

spec/acceptance/hieradata/common.yaml
```diff
@@ -0,0 +1,12 @@
+---
+foreman::client_ssl_ca: /etc/foreman/proxy_ca.pem
+foreman::client_ssl_cert: /etc/foreman/client_cert.pem
+foreman::client_ssl_key: /etc/foreman/client_key.pem
+foreman::initial_location: Default Location
+foreman::initial_organization: Default Organization
```
ekohl, line 6: Do we really care about these? Not setting these results in the Foreman built in defaults being used, which are `Default location` and `Default organization`:
https://github.com/theforeman/foreman/blob/3ff1fd27fa32f7a52ecd995417521b5b91c9a093/db/seeds.d/050-taxonomies.rb#L9

ekohl (approved):

PR #424, opened 2021-07-01
Ensure Artemis client cert exists before tomcat cert-users.properties

ehelms:
The Artemis client DN is calculated as a Deferred object at manifest
apply time. The client certificate needs to exist on disk before
performing this calculation or else the cert-users.properties
file can end up with a blank entry.


ehelms:
I think the tests will show this breaks idempotency but I do not understand why.

ehelms:
Related to https://github.com/theforeman/puppet-certs/pull/367

A better long term solution is probably https://github.com/theforeman/puppet-candlepin/pull/198

PR #425, opened 2021-07-16
Ensure /etc/foreman exists in acceptance test for Candlepin

ehelms:
Results from https://github.com/theforeman/puppet-certs/pull/371


spec/setup_acceptance_node.pp
```diff
@@ -12,6 +12,10 @@
   ensure => present,
 }
 
+file { '/etc/foreman':
+  ensure => directory
+}
```
wbclark, line 17: I am not easily finding information about puppet's defaults for owner, group, and mode with posix provider.... will this be `root:root` and `755` ?

ekohl: I think it's derived from the user running the process then and the umask. In practice this means root:root and probably 755 but possibly 775. It is why I like being explicit in our production code: predictable and explicit.

wbclark: I saw that this is included in https://github.com/theforeman/puppet-katello/pull/423 so we know it works.

I'll leave it up to @ehelms if he thinks it's worth it to be explicit here as well

wbclark (approved):

spec/setup_acceptance_node.pp
ehelms: Given this is just the testing layer, I'm good with this as it is.

PR #427, opened 2021-07-28
Configure Candlepin with CA key password

ehelms:


ehelms:
As the tests should show, this will not protect a deployment if the CA key is unprotected. However, to be more explicit I have a corresponding PR in https://github.com/theforeman/puppet-certs/pull/376 to stop unprotecting the CA key.

wbclark (approved):
This results in the ca_key_password being added to the candlepin.conf file at https://github.com/theforeman/puppet-candlepin/blob/8e4a5aabc7dc2ae1966181e807a924260c2f374f/templates/candlepin.conf.erb#L25-L27

@ehelms does merge order matter between this and https://github.com/theforeman/puppet-certs/pull/376 ?

ehelms:
From my own testing, and from the acceptance tests here, having this configured on Candlepin but with an unprotected key does not seem to break anything. So this can safely go in. We, however, cannot merge puppet-certs first or stand-alone because then Candlepin would break as it would have a protected key and no password supplied.

PR #428, opened 2021-07-28
Use Foreman client certificates to setup qpid queues

ehelms:
Katello uses the Foreman client certificates to talk to Qpid when
katello-agent support is enabled, the same should be used when
setting up Qpid queues and exchanges and not the server certificates
that are intended to be used by Qpid itself.

### DIFF
--- a/manifests/qpid.pp
+++ b/manifests/qpid.pp
@@ -48,9 +48,11 @@ class katello::qpid (
   contain qpid
 
   if $katello::params::enable_katello_agent {
+    include certs::foreman
+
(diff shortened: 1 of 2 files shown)


wbclark (commented):
`qpid_spec.rb` needs updating. otherwise this looks good I think

wbclark (approved):
LGTM.

Please add a Redmine to the tracker at https://projects.theforeman.org/issues/32381 and update the commit message here to include the issue # if you would please before merging.

ehelms:
Previous run was green, failing due to nightlies needing to pass and push with 3.0/3.1 changes in packaging.

PR #430, opened 2021-10-24
Release 21.1.0

ehelms:
Pending release PRs:

- [x] https://github.com/theforeman/puppet-katello/pull/431

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## [21.1.0](https://github.com/theforeman/puppet-katello/tree/21.1.0) (2021-11-05)
+
+[Full Changelog](https://github.com/theforeman/puppet-katello/compare/21.0.0...21.1.0)
+
+**Implemented enhancements:**
+
+- Fixes [\#33830](https://projects.theforeman.org/issues/33830) - Drop content type settings [\#433](https://github.com/theforeman/puppet-katello/pull/433) ([ekohl](https://github.com/ekohl))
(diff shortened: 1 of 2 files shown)


ekohl:
I still think we need to know if this needs modification:
https://github.com/theforeman/puppet-katello/blob/52770f0d2cdcd4ed86b889f4913735ab17a57741/templates/katello.yaml.erb#L14
Either it's redundant and we can drop the whole block or it needs to be configurable now that OSTree is back again.
(guess why I submitted #432)

ehelms:
> I still think we need to know if this needs modification:
> 
> https://github.com/theforeman/puppet-katello/blob/52770f0d2cdcd4ed86b889f4913735ab17a57741/templates/katello.yaml.erb#L14
> 
> 
> Either it's redundant and we can drop the whole block or it needs to be configurable now that OSTree is back again.
> (guess why I submitted #432)

@jlsherrill can you comment on this for us? Is this entry being re-used and is it needed for the new ostree work?

jlsherrill:
oh all of those content_type definitions can go away now !  We figure it out dynamically

ekohl:
Great, since which version?

jlsherrill:
since katello 4.2    https://projects.theforeman.org/issues/31616

ekohl:
I've opened https://github.com/theforeman/puppet-katello/pull/433 and https://github.com/theforeman/foreman-installer/pull/730 to reflect that.

ehelms:
Updated

ekohl (approved):

PR #433, opened 2021-11-02
Fixes #33830 - Drop content type settings

ekohl:
Since Katello 4.2 these are no longer needed and determined dynamically.


ehelms:
Some additional unit test fixes needed :+1: on the simplification

ehelms (approved):

ehelms:
@ekohl do you want to go backwards incompatible with this? or enhancement with README scoping for supported Katello version?

ekohl:
I thought about scoping for the supported Katello version in README. It was unclear to me what we actually do support in this module.

PR #437, opened 2021-12-21
Configure Candlepin with unencrypted CA key

ehelms:
This reverts commit bda1511b88b7954230a0e3bb1a2a71c5b6920e5e.

See https://github.com/theforeman/puppet-certs/pull/386

Candlepin cannot start when a encrypted key is generated using genpkey or on a FIPS enabled machine. Therefore, we should revert to handing Candlepin an unencrypted key.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -20,7 +20,7 @@ fixtures:
       repo: 'https://github.com/puppetlabs/puppetlabs-selinux_core'
       puppet_version: '>= 6.0.0'
     stdlib:        "https://github.com/puppetlabs/puppetlabs-stdlib.git"
-    systemd:       "https://github.com/camptocamp/puppet-systemd.git"
+    systemd:       "https://github.com/voxpupuli/puppet-systemd.git"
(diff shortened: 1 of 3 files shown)


ekohl (commented):
So this will require a bump of the minimum certs module version for it to work?

ehelms:
> So this will require a bump of the minimum certs module version for it to work?

Aye

evgeni:
> So this will require a bump of the minimum certs module version for it to work?

so, which version should we bump `puppet-certs` to?

evgeni:
It's green, and the last time I tested this, it worked ;)

ekohl:
I had hoped we already released a new certs module by now. I'll get on that and then we can modify `metadata.json` to match.

ekohl:
Updated the PR and description. I think this is now ready. Shouldn't be released before https://github.com/theforeman/puppet-certs/pull/391 but our CI uses git master anyway so it doesn't block merge.

ekohl (approved):

PR #439, opened 2022-01-27
enable dnf module for 4.3+

evgeni:


manifests/repo.pp
```diff
@@ -22,4 +22,14 @@
     enabled  => true,
   }
   -> anchor { 'katello::repo': } # lint:ignore:anchor_resource
+
+  if $facts['os']['release']['major'] == '8' and ($repo_version == 'nightly' or versioncmp($repo_version, '4.3') >= 0) {
+    package { 'katello-dnf-module':
+      ensure      => "el${facts['os']['release']['major']}",
```
ekohl, line 28: Should this be
```suggestion
      ensure      => $dist,
```

manifests/repo.pp
```diff
@@ -22,4 +22,14 @@
     enabled  => true,
   }
   -> anchor { 'katello::repo': } # lint:ignore:anchor_resource
+
+  if $facts['os']['release']['major'] == '8' and ($repo_version == 'nightly' or versioncmp($repo_version, '4.3') >= 0) {
```
ekohl, line 26: Should this be on non-EL7 in preparation for EL9?
```suggestion
  if $facts['os']['release']['major'] != '9' and ($repo_version == 'nightly' or versioncmp($repo_version, '4.3') >= 0) {
```

manifests/repo.pp
evgeni: probably, even tho I don't really understand this param… like, why would you override it?

manifests/repo.pp
evgeni: you surely meant `!= 7` and yeah, probably, but I copied this from the `puppet-foreman` module, and that also had `== 8`.

manifests/repo.pp
ekohl: I think it mostly stems from Amazon Linux 2 which reports a major version of 2 but really is el7. This is sort of a compromise: you can override it but we don't provide a decent mapping. There's also no harm in defining it this was rather than as a static variable inside the body.

manifests/repo.pp
ekohl: I indeed mean that and it only occurred to me when reviewing this patch.

manifests/repo.pp
```diff
@@ -22,4 +22,14 @@
...
+    package { 'katello-dnf-module':
+      ensure      => $dist,
+      name        => 'katello',
+      enable_only => true,
+      provider    => 'dnfmodule',
+      require     => Yumrepo['katello'],
```
ekohl, line 32: This probably needs a `before => Anchor['katello::repo']`.

Technical explanation: you can use a collector on a resource (anchor in this case) but not on a class. So in code you can write `Anchor <| title == 'katello::repo' |> -> Package['katello']` and be sure that if `katello::repo` was included, it's done but if it isn't, then it's ignored.

evgeni:
:green_apple:

ekohl (approved):
Thanks!

PR #443, opened 2022-03-31
Fixes #34708 - configure certs for the katello agent broker

evgeni:


evgeni:
the EL7 acceptance tests are failing because of qpid/epel :see_no_evil:

ehelms (approved):
I had originally attempted to avoid specifying defaults to allow the app to determine that but I am can live with this.

ehelms:
> > I had originally attempted to avoid specifying defaults to allow the app to determine that but I am can live with this.
> 
> I mean, it does, just wrongly
> 
> https://github.com/Katello/katello/blob/dfacc81503e66aab70bb2131a25297f50ba611b6/lib/katello/engine.rb#L60-L67

Just the CA certificate.

ehelms:
> Fair. Want me to only set CA? That's what's broken, the rest is just for consistency.

Yes. No. Since user's have no installer options to change these values perhaps being explicit here is OK since it can help with debugging.

ehelms (approved):

PR #450, opened 2022-06-02
Fixes #35005: Drop setting Pulp client certificate settings

ehelms:
These settings were introduced with Pulp 2 and carried over to Pulp
3 before we introduced client authentication support in smart_proxy_pulp.
The support for using Foreman client certificates to talk to Pulp 3 was
introduced in Foreman 3.0. This is now safe to drop.


chris1984 (approved):
I still have green button access on the repo, so I think my review counts :) 

APJ

PR #452, opened 2022-06-08
Fixes #35037 - Add redhat_repository_url to puppet-katello

chris1984:


ehelms:
This seems like an odd runtime configuration, architecturally speaking I am surprised it is not an application Setting a user can configure without needing to run the installer and bounce services. Would that be something you'd consider looking into?

chris1984:
@ehelms I am happy to look at adding it as a setting to Katello. I added it here based on this comment:

https://bugzilla.redhat.com/show_bug.cgi?id=2090466#c7

If you agree we can add it into Katello to get around that issue, I will close this out and open a pr against Katello.

ehelms:
I noticed this is here it is used (https://github.com/Katello/katello/blob/ffda83fef64e878b4d0f8c5d6748278f0fc3a1dc/app/models/katello/cdn_configuration.rb#L49) and pulls from SETTINGS but this could be converted to a Setting (which is an application config vs runtime configuration) and then allow an admin to set it without service restarts and installer runs.

I must admit that the bug itself (https://bugzilla.redhat.com/show_bug.cgi?id=2090466) is not something I understand. This setting (regardless of where it lives) seems to be a work around but not address the root issue that we do not follow redirects properly. I could not grok if that was a design problem or simply a problem we could not solve other than this work around.

ehelms:
We introduce this to fix a users workflow as a work around, if we go that route I would ask for an action item  towards the top of the backlog to make this a Setting or address the general problem with redirects. These kind of runtime settings are not good for users (since it causes service restarts and full installer runs) and it sets us back on forward thinking about HA -- this setting would need to be the same across all instances on disk rather than simply set one place in the database.

chris1984:
Sounds good, will close this one out in favor of a settings pr in Katello. Thanks for the feedback and discussion.

PR #457, opened 2022-09-02
Calculate the client DN for Artemis, ignoring empty values

ehelms:
Moving the computation of the Artemis client DN into this class
instead of relying on the value provided directly by classes in
puppet-certs aims to solve two problems:

 1) Testing at all levels of the value of the client DN
 2) Adding a function to calculate the DN based on a set of values and
    ignoring any empty values which do not end up in the actual certificate

Additionally this removes the layers of indirection `certs::candlepin::artemis_client_dn` which pointed `certs::foreman::client_dn` and moves the calculation closer to the input as this class serves to bring another of items together to properly configure Candlepin/Artemis/Tomcat.

This aims to solve a reported [BZ from Satellite](https://bugzilla.redhat.com/show_bug.cgi?id=1964037) where if a user customizes parts of the certificate creation process, the statically calculated client DN breaks and the Artemis connection breaks. By ignoring empty values the right DN gets provided.

Somewhere in the future, there is a path to eventually calculate this directly from the certificate and avoid even more ambiguity -- my attempts at this previously have proved difficult.

### DIFF
--- a/lib/puppet/functions/katello/build_dn.rb
+++ b/lib/puppet/functions/katello/build_dn.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'yaml'
+# @summary
+#   Convert an array of attribute pairs to a DN string, ignoring empty values
+#
+# @example Pass a set of
+#   $client_dn = katello::build_dn([['CN', 'foo.example.com'], ['O', 'my_org']])
+Puppet::Functions.create_function(:'katello::build_dn') do
+  # @param options
+  #
+  # @return String
+  dispatch :build_dn do
+    param 'Array', :options
+  end
+
+  def build_dn(options)
+    dn = ''
+
+    options.each_with_index do |value, index|
+      next if value[1].nil?
+      dn += "#{options[index][0]}=#{value[1]}"
+      dn += ', ' unless index == (options.length - 1)
+    end
+
+    dn
+  end
+end
--- a/lib/puppet/functions/katello/build_dn.rb
+++ b/lib/puppet/functions/katello/build_dn.rb
@@ -8,21 +8,12 @@ require 'yaml'
 #   $client_dn = katello::build_dn([['CN', 'foo.example.com'], ['O', 'my_org']])
 Puppet::Functions.create_function(:'katello::build_dn') do
   # @param options
-  #
-  # @return String
   dispatch :build_dn do
-    param 'Array', :options
+    param 'Array[Tuple[String[1], Optional[String[1]]]]', :options
+    return_type 'String'
   end
 
   def build_dn(options)
-    dn = ''
-
-    options.each_with_index do |value, index|
-      next if value[1].nil?
-      dn += "#{options[index][0]}=#{value[1]}"
-      dn += ', ' unless index == (options.length - 1)
-    end
-
-    dn
+    options.select { |_key, value| value }.map { |key, value| "#{key}=#{value}" }.join(', ')
   end
 end
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -31,7 +31,10 @@ class katello::candlepin (
   Boolean $manage_db = true,
 ) {
   include certs
+  include certs::foreman
   include katello::params
+
+  $client_dn = katello::build_dn([['CN', $certs::foreman::hostname], ['OU', $certs::foreman::org_unit], ['O', $certs::foreman::org], ['ST', $certs::foreman::state], ['C', $certs::foreman::country]])
 
   class { 'certs::candlepin':
     hostname             => $katello::params::candlepin_host,
@@ -50,7 +53,7 @@ class katello::candlepin (
     keystore_password            => $certs::candlepin::keystore_password,
     truststore_file              => $certs::candlepin::truststore,
     truststore_password          => $certs::candlepin::truststore_password,
-    artemis_client_dn            => $certs::candlepin::artemis_client_dn,
+    artemis_client_dn            => $client_dn,
     java_home                    => '/usr/lib/jvm/jre-11',
     java_package                 => 'java-11-openjdk',
     enable_basic_auth            => false,
--- a/manifests/candlepin.pp
+++ b/manifests/candlepin.pp
@@ -19,6 +19,9 @@
 #   The CA certificate to verify the SSL connection to the database with
 # @param manage_db
 #   Whether to manage the database. Set this to false when using a remote database
+# @param artemis_client_dn
+#   The Distinguished Name of the client certificate that's allowed to access
+#   Artemis. It should still be signed by the correct Certificate Authority.
 class katello::candlepin (
   Stdlib::Host $db_host = 'localhost',
   Optional[Stdlib::Port] $db_port = undef,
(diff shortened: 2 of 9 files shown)


ehelms:
@ekohl could I ask for a round of review on this idea?

lib/puppet/functions/katello/build_dn.rb
```diff
@@ -0,0 +1,28 @@
...
+#   Convert an array of attribute pairs to a DN string, ignoring empty values
+#
+# @example Pass a set of
+#   $client_dn = katello::build_dn([['CN', 'foo.example.com'], ['O', 'my_org']])
+Puppet::Functions.create_function(:'katello::build_dn') do
+  # @param options
+  #
+  # @return String
+  dispatch :build_dn do
+    param 'Array', :options
```
ekohl, line 14: I think this is a stricter validation of what you describe:
```suggestion
    param 'Array[Tuple[String[1], Optional[String[1]]]', :options
```

lib/puppet/functions/katello/build_dn.rb
```diff
@@ -0,0 +1,28 @@
...
+  def build_dn(options)
+    dn = ''
+
+    options.each_with_index do |value, index|
+      next if value[1].nil?
+      dn += "#{options[index][0]}=#{value[1]}"
+      dn += ', ' unless index == (options.length - 1)
+    end
+
+    dn
```
ekohl, line 26: I'd prefer
```suggestion
    options.select { |_key, value| value }.map { |key, value| "#{key}=#{value}" }.join(', ')
```

spec/functions/katello_build_dn_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  it 'should exist' do
+    is_expected.not_to eq(nil)
+  end
+
+  it 'should compute dn' do
+    is_expected.to run.with_params([['a', 1], ['b', 2]]).and_return("a=1, b=2")
+  end
+
+  it 'should compute dn and ignore empty values' do
+    is_expected.to run.with_params([['a', ], ['b', 2]]).and_return("b=2")
```
ekohl, line 13:
```suggestion
    is_expected.to run.with_params([['a', nil], ['b', 2]]).and_return("b=2")
```

manifests/candlepin.pp
```diff
@@ -31,8 +31,11 @@
   Boolean $manage_db = true,
 ) {
   include certs
+  include certs::foreman
```
ekohl, line 34: I don't quite like that this pulls in Foreman certificates, even though they may not be used at all (this class is still intended to be used standalone on a server without Katello).

Instead, I'd suggest to make add `$artemis_client_dn` as a class parameter here. Then in `init.pp` you would pass it. For that to work you need to swap the order of `katello::candlepin` and `katello::application` (due to compile time dependencies, no actual chaining).

You could set `$artemis_client_dn` (as you calculate it below) in `application.pp` and reuse it in `init.pp`. That would signal the relationship.

ekohl (commented):
I like the concept, but I think it breaks some design choices.

The idea is that `katello::candlepin` could be deployed on one server and `katello::application` on another (requiring the user/admin to take care of any relationship between the two). Then there is `katello` for the all-in-one installation we use in our installer. I'll grant you that we may be a long way away from a true split deployment, but this shows how you could build HA setups.

manifests/candlepin.pp
```diff
@@ -50,7 +53,7 @@
     keystore_password            => $certs::candlepin::keystore_password,
     truststore_file              => $certs::candlepin::truststore,
     truststore_password          => $certs::candlepin::truststore_password,
-    artemis_client_dn            => $certs::candlepin::artemis_client_dn,
```
ehelms, line 53: Does the prior existence of this not already imply the behavior you mentioned trying to avoid with this change? This relies on `certs::candlepin` which in turn relies on `certs::foreman` -- https://github.com/theforeman/puppet-certs/blob/master/manifests/candlepin.pp#L27

manifests/candlepin.pp
ehelms: Hrmm, testing that idea out I am running into cyclic dependency issues:

```
        Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Class[Certs::Candlepin] is already declared; cannot redeclare (file: /home/ehelms/workspace/upstream/installer/puppet-katello/spec/fixtures/modules/katello/manifests/candlepin.pp, line: 37) (file: /home/ehelms/workspace/upstream/installer/puppet-katello/spec/fixtures/modules/katello/manifests/candlepin.pp, line: 37, column: 3) on node wareagle.com
```

Thoughts on https://github.com/theforeman/puppet-katello/pull/457#discussion_r983798270 ?

lib/puppet/functions/katello/build_dn.rb
ekohl: Even easier:

```suggestion
    options.compact.map { |key, value| "#{key}=#{value}" }.join(', ')
```

ekohl: Oh, that doesn't work since it's not a hash. Disregard the last suggestion.

lib/puppet/functions/katello/build_dn.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'yaml'
+# @summary
+#   Convert an array of attribute pairs to a DN string, ignoring empty values
+#
+# @example Pass a set of
+#   $client_dn = katello::build_dn([['CN', 'foo.example.com'], ['O', 'my_org']])
+Puppet::Functions.create_function(:'katello::build_dn') do
+  # @param options
+  #
+  # @return String
```
ekohl, line 12: Technically you need to write
```suggestion
  # @return [String] human readable description
```

However, I prefer to use `return_type 'String'` in the dispatch, similar to `param`.

manifests/candlepin.pp
ekohl: Yes, I do think that eventually `certs::candlepin` should no longer include `certs::foreman`. It'll be a breaking change, but worth it.

ehelms:
Currently, acceptance tests are failing due to foreman.service timing out on boot through systemd. I observed this locally and that https://github.com/theforeman/foreman-packaging/pull/8660 seemed to provide a fix.... granted that is not a complete root cause analysis but does close a gap. I could see no other obvious evidence as to why it was suddenly timing out on start yet.

ehelms:
Testing the same assertion round trip, seems to fix test runs locally as well:

```
diff --git a/spec/acceptance/katello_spec.rb b/spec/acceptance/katello_spec.rb
index 98abe98..963422d 100644
--- a/spec/acceptance/katello_spec.rb
+++ b/spec/acceptance/katello_spec.rb
@@ -8,6 +8,8 @@ describe 'Scenario: install katello' do
       <<-PUPPET
         include katello
         include foreman::cli
+
+        package { 'rubygem-sd_notify': ensure => installed }
       PUPPET
     end
   end
```

ekohl (approved):
I rebased it and included the return type. I'm OK with ignoring the sd_notify bit for now.

PR #460, opened 2023-01-27
Allow but filter out empty strings in build_dn

ehelms:
There exists the possibility a parameter passed to build_dn could be an empty string which breaks the previous requirement on a string of length at least one, String[1]. Change to accept any length string but if the string happens to be empty, filter it out of the calculation.

### DIFF
--- a/lib/puppet/functions/katello/build_dn.rb
+++ b/lib/puppet/functions/katello/build_dn.rb
@@ -9,11 +9,12 @@ require 'yaml'
 Puppet::Functions.create_function(:'katello::build_dn') do
   # @param options
   dispatch :build_dn do
-    param 'Array[Tuple[String[1], Optional[String[1]]]]', :options
+    param 'Array[Tuple[String[1], Optional[String]]]', :options
     return_type 'String'
   end
 
   def build_dn(options)
-    options.select { |_key, value| value }.map { |key, value| "#{key}=#{value}" }.join(', ')
+    options_with_values = options.select { |_key, value| !value.nil? && value != '' }
+    options_with_values.map { |key, value| "#{key}=#{value}" }.join(', ')
   end
 end
--- a/spec/functions/katello_build_dn_spec.rb
+++ b/spec/functions/katello_build_dn_spec.rb
@@ -12,4 +12,8 @@ describe 'katello::build_dn' do
   it 'should compute dn and ignore empty values' do
     is_expected.to run.with_params([['a', nil], ['b', '2']]).and_return("b=2")
   end
+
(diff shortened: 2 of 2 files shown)


lib/puppet/functions/katello/build_dn.rb
```diff
@@ -9,11 +9,12 @@
...
   # @param options
   dispatch :build_dn do
-    param 'Array[Tuple[String[1], Optional[String[1]]]]', :options
+    param 'Array[Tuple[String[1], Optional[String]]]', :options
     return_type 'String'
   end
 
   def build_dn(options)
-    options.select { |_key, value| value }.map { |key, value| "#{key}=#{value}" }.join(', ')
+    options_with_values = options.filter { |_key, value| !value.nil? && value != '' }
```
ekohl, line 17: I think this is more expressive in what it tries to achieve:
```suggestion
    options_with_values = options.delete_if { |_key, value| value.nil? || value == '' }
```

I admit Ruby is always a bit weird since there's multiple ways of doing things. For example `filter` is an alias for `select`.

ekohl (approved):
I'm fine either way, let me know if you think my suggestion may be better.

lib/puppet/functions/katello/build_dn.rb
ehelms: I think if we use `delete_if` it becomes:

```
  options.delete_if { |_key, value| value.nil? || value == '' }
```

My understanding reading the docs on it is that it's an in place modifier.

ehelms: And testing, this actually leads to an error:

```
        expected katello::build_dn([["a", nil], ["b", "2"]]) to have returned "b=2" instead of raising FrozenError(can't modify frozen Array: [["a", nil], ["b", "2"]])
        /home/ehelms/workspace/upstream/installer/puppet-katello/spec/fixtures/modules/katello/lib/puppet/functions/katello/build_dn.rb:17:in `delete_if'
```

ekohl: That makes sense, thanks.

Though I think you should use `select` instead of `filter:
```
  Evaluation Error: Error while evaluating a Function Call, undefined method `filter' for #<Array:0x0000561f726d9560> (file: /home/runner/work/puppet-katello/puppet-katello/spec/fixtures/modules/katello/manifests/application.pp, line: 40, column: 24) on node fv-az572-767.xfljd2mj00zetb5gvhv1u2zrtd.dx.internal.cloudapp.net
```

PR #476, opened 2023-09-28
Update gems to support testing Puppet 8

ekohl:


ehelms (approved):

ekohl:
Looks like there is some idempotency issue around the candlepin keystore password file:
```
  Notice: /Stage[main]/Certs::Candlepin/File[/etc/pki/katello/keystore_password-file]/content: content changed '{sha256}3e7d27401b4b9fd9ec4e0ae56ac5bc8e44ec30d17cea992d4ece230ee85234f4' to '{sha256}dab17449f63856b4995ec58958b1df1451af5f6b1e3af904d188e04211ac1fbf'
```

Which is weird, because I'd expect that to pass. Looks like this started to fail last week: https://github.com/theforeman/puppet-katello/actions/runs/6427358289

That's around the time I merged the changes to address the CVE, so now I wonder if I messed that up.

ehelms:
> Looks like there is some idempotency issue around the candlepin keystore password file:
> 
> ```
>   Notice: /Stage[main]/Certs::Candlepin/File[/etc/pki/katello/keystore_password-file]/content: content changed '{sha256}3e7d27401b4b9fd9ec4e0ae56ac5bc8e44ec30d17cea992d4ece230ee85234f4' to '{sha256}dab17449f63856b4995ec58958b1df1451af5f6b1e3af904d188e04211ac1fbf'
> ```
> 
> Which is weird, because I'd expect that to pass. Looks like this started to fail last week: https://github.com/theforeman/puppet-katello/actions/runs/6427358289
> 
> That's around the time I merged the changes to address the CVE, so now I wonder if I messed that up.

I can take a look.

ehelms:
So, I am thinking your change simply uncovered a situation that already existed perhaps? There is a weird aspect I do not understand. On first run of the a manifest that calls `include katello` like our tests do:

```
# ls /opt/puppetlabs/puppet/cache/foreman_cache_data/
admin_password	ca_key_password  candlepin_db_password	candlepin_oauth_secret	db_password  oauth_consumer_key  oauth_consumer_secret
```

Notice that he keystore and truststore password files are missing. On the second application those files will exist.

ekohl:
I found it: `katello` pulls in `foreman-installer-katello` (https://github.com/theforeman/foreman-packaging/blob/8629ae7c79663fca62d289ebff547b0520409c67/packages/katello/katello/katello.spec#L31). We added a migration there that removes the answers (https://github.com/theforeman/foreman-installer/blob/develop/config/katello.migrations/231003142402-reset-store-credentials.rb). So the sequence is:

* Apply `include katello`
* Generates credentials in a cache
* Installer is installed
* Migrations run as part of the RPM installation
* Cache gets removed
* Apply `include katello`
* Generates new credentials in a cache

Result is that the run is not idempotent.

I found this by chance when I used it with the old gems without specifying the Puppet version, which then used Puppet 6. Then the installer refused to install.

ekohl:
I pushed a change that I thought would work, but locally it doesn't appear to help. Still investigating.

ehelms (approved):
Ugly, but it gets the job done.

PR #478, opened 2023-10-04
install katello-selinux when SELinux is enabled

eb4x:
Similar to theforeman/puppet-foreman#79

katello-selinux allows foreman_rails_t to connect to candlepin_activemq_port_t

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -65,6 +65,12 @@ class katello (
     }
   }
 
+  if $facts['os']['selinux']['enabled'] {
+    package { 'katello-selinux':
+      ensure => installed,
+    }
+  }
+
   class { 'katello::application':
     rest_client_timeout => $rest_client_timeout,
     hosts_queue_workers => $hosts_queue_workers,


ekohl (commented):
I was looking at https://github.com/theforeman/foreman-packaging/blob/9295bf93ef55c6f2070f6a96def212796c79ba86/packages/katello/katello/katello.spec#L36 and now I wonder why we don't depend on katello-selinux. I do see rubygem-katello used to depend on katello-selinux, but this feels weird.

@evgeni any idea? Should we fix this in packaging instead?

evgeni:
In the past, adding `Requires: something-selinux` was frowned upon, as it would pull in SELinux stuff on systems that do not use SELinux. However, in the meantime we have rich dependencies and [the guidelines say](https://fedoraproject.org/wiki/SELinux/IndependentPolicy#Adding_dependency_to_the_spec_file_of_corresponding_package):
```
Requires: (%{name}-selinux if selinux-policy-%{selinuxtype})
```

So I guess, it's fine to fix in packaging?

ekohl:
I've opened https://github.com/theforeman/foreman-packaging/pull/9924 for that.

eb4x:
> I've opened [theforeman/foreman-packaging#9924](https://github.com/theforeman/foreman-packaging/pull/9924) for that.

So it'll automagically get installed when katello is installed, and we don't need this right?

Does the same already apply to foreman, and could we strip out the foreman-selinux there aswell? https://github.com/theforeman/puppet-foreman/blob/master/manifests/install.pp#L7-L11

eb4x:
Nm, I see PR adds to both foreman and katello. I must have missed it the first time around :face_with_head_bandage:

ekohl:
> So it'll automagically get installed when katello is installed, and we don't need this right?

I just checked and it's now an RPM dependency:
```console
# dnf install https://yum.theforeman.org/releases/nightly/el8/x86_64/foreman-release.rpm https://yum.theforeman.org/katello/nightly/katello/el8/x86_64/katello-repos-latest.rpm
# dnf repoquery --quiet --whatrequires katello-selinux
rubygem-katello-0:4.11.0-0.4.pre.master.20231114212517git9b51612.el8.noarch
```

So I'm closing this now. Thanks for bringing this to our attention.

PR #480, opened 2023-10-13
Remove traces of EL7 from testing purge_katello

ehelms:


spec/support/acceptance/purge.rb
```diff
@@ -1,4 +1,4 @@
 def purge_katello
-  on default, 'yum -y remove foreman* tfm-*'
+  on default, 'dnf -y remove foreman*'
```
evgeni, line 2: Should this also purge rubygem-*?

ehelms:
```
dnf list installed | grep rubygem
```

None left on the system.

ekohl (approved):
I did a trivial rebase since acceptance was failing and I want to be sure it passes.

PR #481, opened 2023-10-24
allow overriding the name of the metapackage

evgeni:
this is useful for downstream branding

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -54,13 +54,15 @@ class katello (
 
   Integer[0] $hosts_queue_workers = 1,
 ) {
-  package { 'katello':
-    ensure => installed,
-  }
-
   class { 'katello::params':
     candlepin_oauth_key    => $candlepin_oauth_key,
     candlepin_oauth_secret => $candlepin_oauth_secret,
+  }
+
+  if $katello::params::meta_package != '' {
+    package { $katello::params::meta_package:
+      ensure => installed,
+    }
   }
 
   class { 'katello::application':
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -26,4 +26,5 @@ class katello::params (
   Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:${candlepin_port}/candlepin",
   String[1] $candlepin_client_keypair_group = 'foreman',
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
+  String[0] $meta_package = 'katello',
 ) inherits katello::globals {}
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -10,6 +10,7 @@ describe 'katello' do
       it { is_expected.to contain_class('katello::application') }
       it { is_expected.to contain_class('katello::qpid') }
       it { is_expected.to contain_package('rubygem-katello').that_requires('Class[candlepin]') }
+      it { is_expected.to contain_package('katello') }
     end
   end
 end


manifests/init.pp
```diff
@@ -54,15 +54,15 @@
...
-  package { 'katello':
-    ensure => installed,
-  }
-
   class { 'katello::params':
     candlepin_oauth_key    => $candlepin_oauth_key,
     candlepin_oauth_secret => $candlepin_oauth_secret,
   }
 
+  package { $katello::params::meta_package:
```
ekohl, line 62: Can we keep this at the same line as it used to be, to make the diff a bit smaller?

manifests/params.pp
```diff
@@ -26,4 +26,5 @@
   Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:${candlepin_port}/candlepin",
   String[1] $candlepin_client_keypair_group = 'foreman',
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
+  String[1] $meta_package = $katello::globals::meta_package,
```
ekohl, line 29: Do we need to expose this? If we limit it to globals then the installer can set this via Hiera parameters. I don't see a use case (at least in the short term) where the user would want to override this.

It may not be clear today, but it used to have a parameter (until https://github.com/theforeman/puppet-katello/commit/188a71ff52cd3065492b90be73b12f365b021e5a) that was controlled by the installer.

The long term idea was that you have:

* `globals.pp` is set by the installer
* `params.pp` gets exposed to the user via installer parameters
* `application.pp` and `candlepin.pp` are exposed instead of `init.pp` to allow installation of **only** Candlepin or Foreman.

Then we would expose `foreman_proxy_content` as `--content-*` instead.

An intermediate step was https://github.com/theforeman/puppet-katello/pull/209 which introduced manage flags for each component.

Or we can take the easy solution here and inline it. We should do the same with `postgresql_evr_package` which used to have EL7/EL8 logic. When we dropped EL7 support, there's little benefit to keeping it in globals anymore.

ekohl (commented):
Just some random thoughts. I'd like to make the meta package optional. We're not there yet, but perhaps something to consider while doing this. Should we at some point accept `String[0]` to disable this?

manifests/init.pp
evgeni: No, as then the variable is not defined yet, I tried ;)

ekohl: Oh yes, of course.

manifests/params.pp
evgeni: Yeah, I don't particularly care where it's defined as long as it can be overridden via installer/hiera. I'll play around how to minify this.

evgeni: do you have a preference? both ways "work", but I find inlining into `params.pp` nicer as in `init.pp` we never touch `katello::globals` directly, so `package { $katello::globals::meta_package … }` reads odd to me (but works as it's loaded by `katello::params`)

evgeni:
@ekohl the tests fail and I briefly recall this is because of the security fix in candlepin that makes something non idempotent?

ekohl:
Correct. https://github.com/theforeman/puppet-katello/pull/476/commits/3a2be23672b629523a6584f75b36b674f1b5fb1a is the fix for that.

manifests/params.pp
```diff
@@ -26,4 +26,5 @@
   Stdlib::HTTPSUrl $candlepin_url = "https://${candlepin_host}:${candlepin_port}/candlepin",
   String[1] $candlepin_client_keypair_group = 'foreman',
   String[1] $postgresql_evr_package = $katello::globals::postgresql_evr_package,
+  String[0] $meta_package = 'katello',
```
ekohl, line 29: If we make it `String[0]`, I'd prefer to also implement the empty string as "don't install it". Otherwise make it `String[1]`.

ekohl (changes requested):

manifests/init.pp
```diff
@@ -54,15 +54,17 @@
...
-  package { 'katello':
-    ensure => installed,
-  }
-
   class { 'katello::params':
     candlepin_oauth_key    => $candlepin_oauth_key,
     candlepin_oauth_secret => $candlepin_oauth_secret,
   }
 
+  if $katello::params::meta_package {
```
ekohl, line 62: Strings work like in Ruby, not Python. So `''` is still true.
```suggestion
  if $katello::params::meta_package != '' {
```

ekohl (changes requested):

ekohl (approved):

PR #496, opened 2024-05-21
Allow theforeman/foreman 25.x

evgeni:


ekohl (approved):
Argh, I thought I checked this before I did the release.

PR #503, opened 2024-08-05
Fixes #37687 - Uninstall rubygem-anemone on boxes

sjha4:
Ensure rubygem-anemone is removed following removal of the dependency from katello.

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -47,6 +47,10 @@ class katello::application (
     }
   }
 
+  package { 'rubygem-anemone':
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
We have https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/foreman/foreman-obsolete-packages/foreman-obsolete-packages.spec for this.

sjha4:
Oh..Didn't know we had that..Will close this and take it there..

PR #511, opened 2025-02-28
Use the default CA certificate deployed into Foreman directory

ehelms:
Relies on using https://github.com/theforeman/puppet-certs/pull/489

### DIFF
--- a/manifests/application.pp
+++ b/manifests/application.pp
@@ -24,7 +24,7 @@ class katello::application (
   $candlepin_url = $katello::params::candlepin_url
   $candlepin_oauth_key = $katello::params::candlepin_oauth_key
   $candlepin_oauth_secret = $katello::params::candlepin_oauth_secret
-  $candlepin_ca_cert = $certs::ca_cert
+  $candlepin_ca_cert = $certs::foreman::default_ca_cert
   $candlepin_events_ssl_cert = $certs::foreman::client_cert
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Other thought: the intention of https://github.com/Katello/katello/commit/ade73b41f4904dde59131241cc71cfc64ec27e0b was to remove the need for this file to exist at all if the file is the same as what we configure in the `ssl_ca_file` setting, which we [already configure](https://github.com/theforeman/puppet-foreman/blob/d92fea4fb5c0d6dfd453c8b348b0506b3ae67ded/templates/settings.yaml.erb#L20). In other words, can we actually get rid of this setting instead?

Answering my own question: we still need it for the event listener which doesn't look at Foreman's settings:
https://github.com/Katello/katello/blob/d8dc2626709e1303d683585b8ae073b11850b73a/lib/katello/engine.rb#L134-L139

In other words, again the event listener is architecturally so "weird" that it means we need to hack around it. Do you have any thoughts on refactoring Katello to be more in line with other services?

ekohl:
Oh yes, in Katello we configure Smart Proxy with the Server CA, not Default CA. Bleh.

ehelms:
Here is the change to deploy a bundle with both and then deploy that to `/etc/foreman/proxy_ca.pem`. Then we can update the Candlepin code to rely on this and thus only Foreman settings.

PR #522, opened 2026-01-27
Drop support for Puppet 7

evgeni:


metadata.json
```diff
@@ -36,7 +36,7 @@
   "requirements": [
     {
       "name": "puppet",
```
ekohl, line 38: While we're at it, should we change this to `openvox` and drop Puppet support altogether?

ekohl (approved):

metadata.json
evgeni: *change*? not add?

evgeni: add in https://github.com/theforeman/puppet-katello/pull/523

ekohl: I wouldn't mind dropping the unsupported Puppet bits, but don't want to push it.

